PR #50, opened 2012-06-12
DATAGRAPH-190: fixed pagination for findAll()

loki2302:
The fix only works for findAll(). This doesn't do anything to both Cypher and Gremlin query engines, so tests like GraphRepositoryTest::testFindPaged() are slightly misleading: number of pages is always wrong there and this is not checked.


jexp:
Fix was to add a separate count query.

PR #51, opened 2012-06-12
datagraph-146: added a couple of tests for polymorphic behavior

loki2302:
Please let me know if this ticket requires more


jexp:
Merged separately.

Issue #36, opened 2012-02-14
Support for uniqueness


rgan:
Would like to add support for uniqueness using the UniqueFactory functionality in Neo4j 1.6.  An approach is to add a Unique annotation to a field in a node entity which will create the node using the UniqueFactory in EntityStateHandler and NodeEntityState - as implemented in these 3 commits:

https://github.com/rgan/spring-data-neo4j/commit/492929e64676ac58b31c84351aab618dd958faa2
https://github.com/rgan/spring-data-neo4j/commit/ccd230b9e82051141ca2426fb4e91a8b4943173c
https://github.com/rgan/spring-data-neo4j/commit/8fef84e0024235e3dad08b70ae2767f251137e06

Do you see issues with this approach?

PR #90, opened 2012-10-30
spatial #36 - Possible Solution

benday280412:
Hi,

This is part of a possible solution to issue https://github.com/neo4j/spatial/issues/36 raised against neo4j spatial.  There is a corresponding change in spatial for which I will also raise a pull request.

The change tries to use the same approach as used by the Lucene index, which uses a CatchingIteratorWrapper for returning results.  This means that deletion only happens when the geom node comes back as part of a query and is found to lack its corresponding domain node.

I've put some comments and questions in the code where I'm not 100% if my understanding is correct.

I look forward to your feedback.

Cheers,

Ben


jexp:
Sorry, not sure if this is still relevant?

PR #142, opened 2013-11-15
Update applicationContext.xml

gcolbert:
Replaced schema location from :
http://www.springframework.org/schema/data/neo4j/spring-neo4j.xsd
(404 error)
to http://www.springframework.org/schema/data/neo4j/spring-neo4j-2.2.xsd


jexp:
the non-versioned url resolves to an xsd

PR #170, opened 2014-04-04
Changed transaction.close() to transaction.finish() in the cineasts example (compilation error).

skinny85:


jexp:
Should not be relevant anymore. Thanks anyway.

PR #183, opened 2014-05-10
DATAGRAPH-420: Added "ignore case" support. Negation with "ignore case" currently not possible.

BHSDuncan:
This should address the implementation of "IgnoreCase" for repositories; however, I'm not sure Neo4j supports lookaheads in order to implement negation (i.e. for the "Not" cases in the repo interface methods). Currently, the "not" portion of "ignore case" will behave just like the normal "ignore case".  Of course I'm happy to change this if I'm off.

The change also includes various additions to the unit tests.

This is my first pull request for SDN and I'm eager to make any worthwhile, quality contributions I have so please pick my implementation apart and let me know what you'd like to see done differently, if anything.


jexp:
Do we want to revisit this rebased on the latest code-base or is it going to be irrelevant with the upcoming SDN 4. And Neo4j 2.3 will add ILIKE afaik.

PR #189, opened 2014-06-08
DATAGRAPH-257 QueryResultBuilder cannot be cast to java.util.Set

nickithewatt:
@jexp - this replaces the previous outdated PR https://github.com/spring-projects/spring-data-neo4j/pull/130.

Fixes QueryResultBuilder cannot be cast to java.util.Set in NodeEntity for SDN 3.1+ context. Needed to do a bit of a workaround for transitive &#64;Query annotated fields, as these were never being evaluated. I think a previous change to how transitive fields worked rendered &#64;Query fields null unless the transitive modifier was removed, however transitive modifier should be allowed so that it doesn't get serialized.


jexp:
I think we merged it manually, or? Do you remember @nickithewatt ?

PR #230, opened 2015-02-09
Default TypeSafetyPolicy in Neo4jConfiguration

dalsh:
Having repositories return different entities type by default without type safety is not very intuitive. Can you consider changing the default behavior in Neo4jConfiguration as follow ?

Example : 
```Java
@NodeEntity
public class Dog { 
    @GraphId Long nodeId
}

@NodeEntity
public class Cat { 
    @GraphId Long nodeId
}

public interface CatRepository extends GraphRepository<Cat> {}

public interface DogRepository extends GraphRepository<Dog> {}

catRepository.save(new Cat());
dogRepository.save(new Dog());

GET /dogs/0 // returns a Cat
GET /dogs/1 // returns a Dog
```
I might be doing something wrong but I "fixed" this odd behavior by declaring a TypeSafetyPolicy with THROWS_EXCEPTION option in my own configuration, and thought this might be a reasonable default.

### DIFF
--- a/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/config/Neo4jConfiguration.java
+++ b/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/config/Neo4jConfiguration.java
@@ -48,6 +48,7 @@ import org.springframework.data.neo4j.support.relationship.RelationshipEntitySta
 import org.springframework.data.neo4j.support.schema.SchemaIndexProvider;
 import org.springframework.data.neo4j.support.typerepresentation.ClassValueTypeInformationMapper;
 import org.springframework.data.neo4j.support.typerepresentation.TypeRepresentationStrategyFactory;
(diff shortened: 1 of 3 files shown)


jexp:
We didn't change the default due to backwards compatibility. But might make sense to change it in 3.3.x

Would you mind also adapting the docs for this too?

dalsh:
I don't mind adapting the docs at all, except I'm not sure which part of it should be changed. I'll be searching but if you can point me to it I'll have a look right away

jexp:
Darn, it seems the person who added it back then didn't add a section in the docs (it would go into the type representation section, in `src/main/asciidoc/reference/programming-model/typerepresentationstrategy.adoc`

dalsh:
I've updated the PR with the related docs updates. Please feel free to comment and I'll apply fixes as requested.

jexp:
Amazing, thanks so much !

jexp:
Hmm this causes a lot of tests to fail :( Looking into those now.

PR #231, opened 2015-02-14
Documentation polishing

fpavageau:
I've tried to use the code style (for class names) a bit more consistently across the documentation. I've also fixed spelling and grammar issues, mainly after running the documentation through `aspell`.

### DIFF
--- a/src/docbkx/reference/heroku.xml
+++ b/src/docbkx/reference/heroku.xml
@@ -4,7 +4,7 @@
   <title>Heroku: Seeding the Cloud</title>
   <para>
     Deploying your application into the cloud is a great way to scale from from "wouldn't it be cool if.." to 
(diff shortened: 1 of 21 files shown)


jexp:
Thanks a lot for your work, but I'm soo sorry to tell you that we moved to asciidoc as source for our documentation last summer. Just kept the old XML around for reference, we should probably have deleted it.

Please next time before starting a contribution like this, reach out to us for feedback on the idea.

Perhaps you can apply the diff of your changes there?

See: https://github.com/spring-projects/spring-data-neo4j/tree/master/src/main/asciidoc

fpavageau:
Right :)

No problem, I'll have a look at the asciidoc sources and create another PR when I'm done.

PR #277, opened 2015-04-20
Fixed build issues by removing the redundant annotations per the directi...

LordHumungous:
...on of SPR-12245.


jexp:
Thanks a lot, solved separately.

PR #295, opened 2015-06-04
Add paginationsupport for custom queries

Richterrettich:
In order to support paginationsupport for custom repositories, we need a
lower level support for pagination in the session, as well as the
corresponding delegates. This commit will refactor the query augmentation in ParameterisedStatement.java.

related to https://jira.spring.io/browse/DATAGRAPH-653

### DIFF
--- a/neo4j-ogm/src/main/java/org/neo4j/ogm/cypher/statement/ParameterisedStatement.java
+++ b/neo4j-ogm/src/main/java/org/neo4j/ogm/cypher/statement/ParameterisedStatement.java
@@ -17,6 +17,7 @@ import org.neo4j.ogm.cypher.query.Pagination;
 import org.neo4j.ogm.cypher.query.SortOrder;
 
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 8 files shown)


vince-bickers:
Hi Rene

Many thanks for this! We will add this contribution to the code scheduled for the next release because RC1 is in code freeze and no further commits are permitted. 

Subsequent to the RC1 code-freeze the project modules have been being split, and the OGM part is now hosted here: https://github.com/neo4j/neo4j-ogm

The Spring Data integration parts are being retained in the Spring repositories, and the new "master" branch post-split is 4.0-GA. We had always intended to keep these modules separate so that both Spring and non-Spring dev shops could use the OGM,  but for various operational reasons, it wasn't possible to do this for 4.0 M1

Consequently we can't take your pull request directly from this repo, because they apply to the OGM part. Instead I'll patch your changes into the OGM repo today.

Richterrettich:
Hi Vince,

I'm really sorry. I knew (from your SO reply) that 4.0 is frozen right now, but i didn't really know where to fit my changes into. Well, now I know. Thanks for the insight. If you want, I can patch it myself and create a sparate PR für the ogm-package.

I want to complete the paging thing if that is okay. Then 4.0-GA would be the base branch for those changes, right?

vince-bickers:
Lets get in touch so we can discuss how best to proceed.

email: vince@grapahaware.com
skype: vince.bickers

jexp:
I think you two resolved this separately, closing.

PR #303, opened 2015-07-08
Adding tests for incorrect relationship repository queries

nd0:
Hi Luanne,

I  think the genrated query for the repository method 
```
List<Rating> findByUserNameAndMovieName(String userName, String movieName)
```
should be
```
{"statements":[{"statement":"MATCH (n:`User`) WHERE n.`name` = { `userName` } MATCH (m:`Movie`) WHERE m.`name` = { `movieName` } MATCH (n)-[r:`RATED`]->(m) WITH n,r MATCH p=(n)-[*0..1]-() RETURN collect(distinct p), ID(r)","parameters":{"userName":"Michal","movieName":"Speed"},"resultDataContents":["graph","row"],"includeStats":false}]}
```
but the following were send to the neo4j server
```
{"statements":[{"statement":"MATCH (n:`User`) WHERE n.`name` = { `name` } MATCH (m:`Movie`) WHERE m.`name` = { `name` } MATCH (n)-[r:`RATED`]->(m) WITH n,r MATCH p=(n)-[*0..1]-() RETURN collect(distinct p), ID(r)","parameters":{"name":"Speed"},"resultDataContents":["graph","row"],"includeStats":false}]}
```

The second test is not so clear.
If Rating has a method getUserLevel() and RatingRepository a method List<Rating> findByUserLevel(String userLevel) but no attribute userLevel and no setter, then I would expect, that the query is 
```
{"statements":[{"statement":"MATCH (n)-[r:`RATED`]->(m) WHERE n.`level` = { `userLevel` } WITH n,r MATCH p=(n)-[*0..1]-() RETURN collect(distinct p), ID(r)","parameters":{"userLevel":"5"},"resultDataContents":["graph","row"],"includeStats":false}]}
```
and not
```
{"statements":[{"statement":"MATCH (n)-[r:`RATED`]->(m) WHERE r.`userLevel` = { `userLevel` } WITH n,r MATCH p=(n)-[*0..1]-() RETURN collect(distinct p), ID(r)","parameters":{"userLevel":"5"},"resultDataContents":["graph","row"],"includeStats":false}]}
```
but in any case should the should the upper query be send, if I add the annotaion @Transient to the method.

### DIFF
--- a/spring-data-neo4j/src/test/java/org/springframework/data/neo4j/moreQueries/DerivedRelationshipEntityQueryTest2.java
+++ b/spring-data-neo4j/src/test/java/org/springframework/data/neo4j/moreQueries/DerivedRelationshipEntityQueryTest2.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c)  [2011-2015] "Pivotal Software, Inc." / "Neo Technology" / "Graph Aware Ltd."
+ *
+ * This product is licensed to you under the Apache License, Version 2.0 (the "License").
+ * You may not use this product except in compliance with the License.
+ *
+ * This product may include a number of subcomponents with
+ * separate copyright notices and license terms. Your use of the source
+ * code for these subcomponents is subject to the terms and
+ * conditions of the subcomponent's license, as noted in the LICENSE file.
+ */
+
+package org.springframework.data.neo4j.moreQueries;
+
+import org.junit.After;
+import org.junit.Before;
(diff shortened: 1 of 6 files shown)


luanne:
Instead of creating another set of domain objects, I've updated the existing ones. The second test is not supported at the moment and is tracked by https://jira.spring.io/browse/DATAGRAPH-708

PR #320, opened 2015-12-10
Update ExecutingRestRequest to allow HTTPS requests

vdatcu:
Can we have this change implemented, Michael? We are using Neo4j behind an SSL protected nginx and the ExecutingRestRequest will only check for "http://" starting URLs. Otherwise, the url gets modified in bad way, i.e.

**/db/data/https://${domain}.com/db/data/transaction/25**

instead of **/db/data/transaction/25**,

whenever a transaction is created. So the initial transaction is OK, but when subsequent are to be created, requests will be made to ```this.baseUri + "/" + path;```, instead of ```path```

If you need a detailed scenario or nginx log, please let me know.

@jexp

### DIFF
--- a/spring-data-neo4j-rest/src/main/java/org/neo4j/rest/graphdb/ExecutingRestRequest.java
+++ b/spring-data-neo4j-rest/src/main/java/org/neo4j/rest/graphdb/ExecutingRestRequest.java
@@ -104,7 +104,7 @@ public class ExecutingRestRequest implements RestRequest {
     }
 
     private String pathOrAbsolute( String path ) {
-        if ( path.startsWith( "http://" ) ) return path;
+        if ( path.startsWith( "http://" ) || path.startsWith( "https://" ) return path;
         return baseUri + "/" + path;
     }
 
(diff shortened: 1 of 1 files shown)


jexp:
Would you mind:

* adding a test
* merging the two commits into one ?

vlad-datcu-kalon:
Hi Michael, I will do as requested as soon as I get to the office, probably on Monday. Thanks!

vlad-datcu-kalon:
@jexp , there is an issue building the project. I checked out the 3.4.x branch and when I build, this error pops up:

```ERROR] Failed to execute goal on project spring-data-neo4j: Could not resolve dependencies for project org.springframework.data:spring-data-neo4j:jar:3.4.2.BUILD-SNAPSHOT: The following artifacts could not be resolved: com.mysema.querydsl:querydsl-lucene3:jar:4.0.7, com.mysema.querydsl:querydsl-apt:jar:4.0.7: Could not find artifact com.mysema.querydsl:querydsl-lucene3:jar:4.0.7 in spring-libs-snapshot (https://repo.spring.io/libs-snapshot) -> [Help 1]```

The artifact is there, I manually checked. I also tried purging the local repo, deleting the "repository" directory and running the mvn command with -U. Any suggestions?

vlad-datcu-kalon:
I think there is a dependency issue, because as far as I can see, the project is configure to use com.mysema.querydsl:querydsl-dsl/lucene3:4.0.7, and this version has its group id changed to query.dsl. See ```https://repo.spring.io/libs-snapshot/com/mysema/querydsl/querydsl-lucene3/``` The latest version is 3.7.0. Is this the problem?

vlad-datcu-kalon:
@jexp any news?

vlad-datcu-kalon:
Also, my fork of the 3.4.x does not compile :)

mangrish:
Closing. SDN 3.x is no longer actively developed.

PR #321, opened 2015-12-15
Set Neo4j 2.3 as Default and Upgrading QueryDSL to 4.0.7

bradnussbaum:
Adding support for querydsl 4.0.7 upgrade
Upgrading neo4j to 2.3.1 and spatial to 0.15
Setting Neo4j 2.3 as the default version


jexp:
Thanks a lot.
MERGED Manually

PR #328, opened 2016-01-23
Cypher-based Message Source

espiegelberg:
The Spring framework provides the MessageSource interface for resolving messages, with support for the parameterization and internationalization. Two out-of-the-box implementations are provided: ResourceBundleMessageSource, built on top of the standard ResourceBundle, and ReloadableResourceBundleMessageSource, being able to reload message definitions without restarting the VM.

While these implementations work very well, they rely on the message definitions being defined on the filesystem. It is not uncommon for some developers to create a custom MessageSource implementation that is backed by a database, rather than filesystem, and the web has numerous examples of such implementations using JDBC.

I believe a MessageSource implementation that makes use of Cypher to load messages from Neo4j is a worthwhile addition to the Spring Data Neo4j  project. As developed on my SDN fork on github, I have created an implementation (and tests) and am submitting it for contribution.


spring-data-neo4j/src/main/java/org/springframework/data/neo4j/context/support/CypherMessageSource.java
```diff
@@ -0,0 +1,162 @@
...
+                    Map<String, String> nodeProperties = (Map<String, String>) node.getValue();
+
+                    String code = nodeProperties.remove("code");
+
+                    for (Entry<String, String> nodeEntry : nodeProperties.entrySet()) {
+
+                        String text = nodeEntry.getValue();
+
+                        String localeString = nodeEntry.getKey();
+                        Locale locale = LocaleUtils.toLocale(localeString);
```
jexp: probably only look up once per locale then cache?

jexp:
Thanks a lot for your contribution @espiegelberg (have you signed the Neo4j and SpringSource CLAs?)

Really interesting approach, would you mind adding something similar to SDN 4? (ie. branch 4.1.x?)

There is no integration test for actually using the database to populate the message source and no test for overriding the query.

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/context/support/CypherMessageSource.java
espiegelberg: Are you referring to the look up in this particular block of code or the look up of messages through the message source?

While anything that improves performance --- such as caching -- is a good thing, the above block of code is only a one time operation executed when the CypherMessageSource bean is initialized. Given that, is implementing caching here actually beneficial?

That said, if you can provide more context as to what you'd like to see I'll be happy to implement it.

espiegelberg:
Thanks for the feedback @jexp; I'm excited the contribution may work its way into the repo.

I'll follow up about the CLA with you in email.

I would be happy to add something similar to SDN 4. My original assumption was that it could simply be merged from the master to the 4.1.x branch but I now see that there may be some differences of how the integration tests for each branch will be written. I propose that we get the pull request for the master tightened up and accepted and then I'll turn around and create the similar pull request for the 4.1.x branch.

I'll create an integration test that covers populating the message source as well as overriding the query.

espiegelberg:
I've beefed up the test coverage, per your feedback @jexp.

The head of the master is not currently building successfully (see https://jira.spring.io/browse/DATAGRAPH-811), so in the meantime I coded, committed, and pushed my changes and these updates to the 3.4.x branch. Please review and let me know next steps. Because this is now on the 3.4.x branch, should I close this and create a new pull request?

Once we've got things squared away on the 3.4.x branch I'll be happy to create separate contributions and pull requests for the master as well as the 4.1.x branch. (Each branch has slight coding changes required, so I assume that's the desired approach.)

jexp:
can you update the PR too? Right now you only did it locally in your fork.

jexp:
Please note that I'm going to switch the master for the SDN repo to SDN 4.1
and this master will become 3.5.x

jexp:
@espiegelberg I didn't close it it was probably automatically closed when I pushed the branch change. Can you reissue it against branch 3.5.x ? thanks a lot.

espiegelberg:
This pull request has been reissued against the 3.5.x branch as https://github.com/spring-projects/spring-data-neo4j/pull/329

PR #329, opened 2016-01-26
Cypher-based Message Source on 3.5.x branch

espiegelberg:
Resubmitted against the newly created 3.5.x branch.

The previous submission was against the 3.4.x branch and was pull request https://github.com/spring-projects/spring-data-neo4j/pull/328

------------------------

The Spring framework provides the MessageSource interface for resolving messages, with support for the parameterization and internationalization. Two out-of-the-box implementations are provided: ResourceBundleMessageSource, built on top of the standard ResourceBundle, and ReloadableResourceBundleMessageSource, being able to reload message definitions without restarting the VM.

While these implementations work very well, they rely on the message definitions being defined on the filesystem. It is not uncommon for some developers to create a custom MessageSource implementation that is backed by a database, rather than filesystem, and the web has numerous examples of such implementations using JDBC.

I believe a MessageSource implementation that makes use of Cypher to load messages from Neo4j is a worthwhile addition to the Spring Data Neo4j project. As developed on my SDN fork on github, I have created an implementation (and tests) and am submitting it for contribution.

### DIFF
--- a/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/context/support/CypherMessageSource.java
+++ b/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/context/support/CypherMessageSource.java
@@ -1,0 +1,184 @@
+/**
+ * Copyright 2011-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


mangrish:
3.5 is no longer being actively developed. I'm closing this PR as it can't be integrated into 4.x

PR #334, opened 2016-03-31
Change info to debug level for session init

tinnou:
In a web environment `getSession` can potentially get called very often.


mangrish:
This PR has been fixed in #363 . Closing.

PR #346, opened 2016-06-15
DATAGRAPH-879 - Add repository support using CDI

mp911de:
Spring Data Neo4j now supports CDI by exporting Neo4j repositories as CDI beans. CDI users need to provide a `org.neo4j.ogm.session.Session` bean to connect to Neo4j. The CDI extension supports qualifiers and custom repository implementations.

### DIFF
--- a/spring-data-neo4j/pom.xml
+++ b/spring-data-neo4j/pom.xml
@@ -108,6 +108,28 @@
                 <version>3.1.0</version>
             </dependency>
 
+            <dependency>
+                <groupId>javax.el</groupId>
+                <artifactId>el-api</artifactId>
+                <version>${cdi}</version>
+                <scope>test</scope>
+            </dependency>
+
+            <!-- CDI -->
+            <dependency>
+                <groupId>javax.enterprise</groupId>
+                <artifactId>cdi-api</artifactId>
+                <version>${cdi}</version>
+                <scope>provided</scope>
+                <optional>true</optional>
+            </dependency>
+
+            <dependency>
+                <groupId>org.apache.openwebbeans.test</groupId>
+                <artifactId>cditest-owb</artifactId>
+                <version>${webbeans}</version>
+                <scope>test</scope>
+            </dependency>
         </dependencies>
 
     <build>
--- a/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/cdi/Neo4jCdiRepositoryBean.java
+++ b/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/cdi/Neo4jCdiRepositoryBean.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.data.neo4j.repository.cdi;
+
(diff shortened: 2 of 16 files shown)


mangrish:
@mp911de I tried to merge this PR today and even though it works on all our development environments it breaks on all of our build machines. It was not clear what exactly was causing several of our tests to fail (see https://build.spring.io/browse/SPRINGDATA-DATAGRAPH-1332) but I have had to revert this and I may hold off on this merge until we discuss the larger scope of the changes scheduled for SDN later this week.

mp911de:
Thanks for your feedback. I wondered whether it's necessary to perform a connection/driver cleanup. 
[Neo4jCdiProducer](https://github.com/spring-projects/spring-data-neo4j/pull/346/files#diff-2df7e7be3b108a4a6e54c47fe6920228R38) creates a `SessionFactory` and a `Session` but I've not seen any `close()`/`shutdown()` method. Feel free to ping me (Skype/Gitter/Slack/...).

mangrish:
Code has been merged but the Integration Test breaks the build. It seems that the CDI Container does some invasive change to the Jersey instance that runs for each integration test which causes subsequent tests that run after this one to dail.  I have commented this out on master - https://github.com/spring-projects/spring-data-neo4j/blob/master/spring-data-neo4j/src/test/java/org/springframework/data/neo4j/repository/cdi/CdiExtensionIT.java.

Could you please look in to this test otherwise we might have to revert this code out again 👎

mp911de:
Do you have pointer for me how to reproduce the test issue?

mangrish:
Heh.. well it's funny you say that. What we do is replicate the way the build servers are set up. So what you can do is pop over to AWS and create a new micro EC2 instance in Ireland (eu-west-1) and use a user created AMI (ami-a4d44ed7). Then connect to the box and do:

-`apt-get update`
-`apt-get upgrade`
- `apt-get  install openjdk-8-jdk maven`

You can then git clone the spring-data-neo4j repository.

From here you run: `mvn clean verify` and the Integration Tests will fail. Make sure you uncomment the CdiExtentionIT to see the failures. Don't be deceived by the output ordering. It's definitely that test!

Something in your test seems to hang around causing subsequent tests to fail. It looks to be caused by something jersey related. I've tried removing the system property in the @AfterClass to no effect.

Good luck!

mp911de:
Thanks for the details. I wasn't able to reproduce the test issue 😢 but I pushed a commit to [mp911de/spring-data-neo4j@issue/DATAGRAPH-879-CDI-Test](https://github.com/mp911de/spring-data-neo4j/tree/issue/DATAGRAPH-879-CDI-Test). That change filters CDI extensions. Therefore, the Jersey CDI extension is not activated anymore.

Could you apply https://github.com/mp911de/spring-data-neo4j/commit/353b167d01a1978410f2834ed38edd72d8aef9d6 to your code (not necessarily master) and run your build on that? If that works, I can submit a PR.

mp911de:
@mangrish ping

mangrish:
@mp911de Sorry was on holiday for the last 10 days. Will try this out and let you know how it goes.

mangrish:
@mp911de I tried it out and unfortunately it still fails. It's pretty easy to get the EC2 instance up. Takes about 5-10 minutes to get something running. I'll open a bug for this tomorrow.

mp911de:
I was finally able to narrow down the issue, and I have found a solution. The reason why the tests weren't reproducible on my machine is that the test order is different. Tests in `org.springframework.data.neo4j.queries` were executed before the CDI test was executed.

Now the reason for the failure: The CDI framework scans all class files and by doing so it loads each class. As `org.springframework.data.neo4j.queries.QueryIntegrationIT` et al. perform a static initialization (`private static GraphDatabaseService graphDatabaseService = getGraphDatabaseService();`) the database service points to the instance initialized in the CDI test. Shutting down the CDI test also shuts down the referenced instance and that's the reason why tests in `org.springframework.data.neo4j.queries` break.

The solution: Remove the static initialization and replace it by a cleaner patter:

```java
	@BeforeClass
	public static void beforeClass(){
		  graphDatabaseService = getGraphDatabaseService();
	}
```

I've done it already in my repo. I can submit the changes as PR.

mangrish:
You're right it worked! https://jira.spring.io/browse/DATAGRAPH-903

PR #389, opened 2017-03-06
DATAGRAPH-951 - Repository finders results get mixed when run concurr…

nmervaillie:
…ently.

Refactor to make things thread safe. Moved filters creation logic from CypherFilter class to dedicated builders.

Commit 917395 introduced the CypherFilter class (sdn version) to wrap ogm Filters in the goal of making things thread safe.
CypherFilter builds dynamically new instances of Filters

Commit f43d76 added support more operators of spring data but introduced FilterFunction in the state of CypherQuery, making it not thread safe anymore

- [X] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [X] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATAGRAPH).
- [X] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [X] You submit test cases (unit or integration tests) that back your changes.


spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/query/derived/builder/DistanceComparisonBuilder.java
```diff
@@ -14,38 +14,71 @@
...
 		super(part, booleanOperator, entityType);
 	}
 
 	@Override
-	public List<CypherFilter> build() {
-		List<CypherFilter> filters = new ArrayList<>();
+	public List<Filter> build(Stack<Object> params) {
 
-		CypherFilter filter = new CypherFilter();
+		Object firstArg = params.pop();
```
vince-bickers: Is this right? If we push 1, 2 we need to pop 2, 1

nmervaillie:
I think this works :the args are pushed in reverse order in the stack
@see org.springframework.data.neo4j.repository.query.derived.CypherFinderQuery#getFilters
Covered by tests
in org.springframework.data.neo4j.examples.restaurants.RestaurantIT

mangrish:
Looks good to me. We will still have to revisit part of this code again in the 5.0 branch when we fix up derived query finder dispatching but i think the structure is fine.

PR #396, opened 2017-07-04
4.1.x

xinan711456:
Hello, I am building a javaweb project. using springmvc+spring+spring data neo4j v4.1.11. something get wrong when I run Tomcat8 that "public interface ActorRepository extends GraphRepository<Actor> can not be autowired".

According this website"http://docs.spring.io/spring-data/data-neo4j/docs/4.1.11.RELEASE/reference/html/",  I run springTest successful but fail in the javaweb mode.

Thanks.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -58,7 +58,7 @@ It supports:
 * automatic mapping annotated domain entities for nodes and relationships
 * interface based repositories with provided, derived, and annotated finder methods
 * transaction control
-* multi-transport (embedded, http, [bolt])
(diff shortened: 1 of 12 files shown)


schauder:
Hi, @xinan711456 what you write sounds more like a support request than a PR. If this is the case please post a question over at StackOverflow, tagged with `spring-data-neo4j`. Feel free to put a link to that question here.

If it is actually a bug or a feature request, please post it in Jira https://jira.spring.io/browse/DATAGRAPH

PR #402, opened 2017-10-16
Fixed a dead link to Spring Transaction Management

aldrinm:
Updated the dead link to Spring Transaction Management with (what I think) is current.


nmervaillie:
Thanks Aldrin !

PR #403, opened 2017-10-18
Improve keywords to query documentation

meistermeier:
Add a table to explain the keywords to cypher query by example.


nmervaillie:
merged

PR #405, opened 2018-01-12
Add support for persistence constructors

nmervaillie:
See https://jira.spring.io/browse/DATAGRAPH-1056

This PR is designed to work with this OGM branch : https://github.com/neo4j/neo4j-ogm/tree/feature/persistence-ctors
tracked by https://github.com/neo4j/neo4j-ogm/pull/452

- [X] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [X] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATAGRAPH).
- [X] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [X] You submit test cases (unit or integration tests) that back your changes.

### DIFF
--- a/spring-data-neo4j/pom.xml
+++ b/spring-data-neo4j/pom.xml
@@ -194,6 +194,80 @@
             <artifactId>spring-webmvc</artifactId>
             <scope>test</scope>
         </dependency>
+		<dependency>
+			<groupId>org.jetbrains.kotlin</groupId>
+			<artifactId>kotlin-stdlib-jdk8</artifactId>
+			<version>${kotlin.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.jetbrains.kotlin</groupId>
+			<artifactId>kotlin-reflect</artifactId>
+			<version>${kotlin.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.jetbrains.kotlin</groupId>
+			<artifactId>kotlin-test</artifactId>
+			<version>${kotlin.version}</version>
+			<scope>test</scope>
+		</dependency>
 
-    </dependencies>
+	</dependencies>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.jetbrains.kotlin</groupId>
+				<artifactId>kotlin-maven-plugin</artifactId>
+				<version>${kotlin.version}</version>
+				<executions>
+					<execution>
+						<id>compile</id>
+						<phase>compile</phase>
+						<goals>
+							<goal>compile</goal>
+						</goals>
+						<configuration>
+							<sourceDirs>
+								<source>src/main/java</source>
+							</sourceDirs>
+						</configuration>
+					</execution>
+					<execution>
+						<id>test-compile</id>
+						<phase>test-compile</phase>
+						<goals>
+							<goal>test-compile</goal>
+						</goals>
+						<configuration>
+							<sourceDirs>
+								<source>src/test/java</source>
+								<source>src/test/kotlin</source>
+							</sourceDirs>
+						</configuration>
+					</execution>
+				</executions>
+				<configuration>
+					<jvmTarget>1.8</jvmTarget>
+				</configuration>
+			</plugin>
+			<plugin>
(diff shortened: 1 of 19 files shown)


spring-data-neo4j/pom.xml
```diff
@@ -194,6 +194,80 @@
...
+			<scope>test</scope>
+		</dependency>
+
+	</dependencies>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.jetbrains.kotlin</groupId>
+				<artifactId>kotlin-maven-plugin</artifactId>
```
frant-hartm: 1. Do we now compile everything (also java) by the kotlin compiler?
2. Is that safe? 
3. Is this how other SD projects do it?

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/config/Neo4jMappingContextFactoryBean.java
```diff
@@ -52,16 +59,28 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
 		Neo4jMappingContext context = new Neo4jMappingContext(sessionFactory.metaData());
 		context.initialize();
 
+		ConversionService conversionService = null;
+		try {
+			conversionService = beanFactory.getBean(ConversionService.class);
+		} catch (NoSuchBeanDefinitionException e) {
+			LOG.debug("Unable to find a conversion service to use for entity instantiation, using none");
+		}
+		sessionFactory.setEntityInstantiator(getEntityInstantiator(context, conversionService));
```
frant-hartm: `getEntityInstantiator(context, conversionService)` may be called with `conversionService = null` - is this ok?

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/config/OgmEntityInstantiatorAdapter.java
```diff
@@ -0,0 +1,88 @@
...
+			Assert.notNull(propertyValues, "Properties cannot be null");
+			this.propertyValues = propertyValues;
+		}
+
+		@SuppressWarnings({"unchecked"})
+		@Override
+		@Nullable
+		public Object getParameterValue(PreferredConstructor.Parameter parameter) {
+			Object value = propertyValues.get(parameter.getName());
+			if (value == null || conversionService == null) {
```
frant-hartm: I guess this answers my previous question.

src/main/asciidoc/reference/neo4j-repositories.adoc
```diff
@@ -663,5 +663,41 @@ Delete graph entities::
...
+
+It means that SDN will try to find the best constructor to instantiate entities.
+Constructor discovery works as follow :
+
+- look for explicit @PersistenceConstructor annotation on constructor
+- if not found, look for the default constructor
+- if not found, look at any specific constructor defined by the entity.
+
+Values for constructor parameters are determined by looking at parameters names (not parameter types).
+For values to be filled in correctly, the parameter names have to match the entity attributes names.
```
frant-hartm: 1. Doesn't this mean domain classes need to be compiled with `-parameters` flag? Should be mentioned here.
1. How does the test code end up with it? I don't see it anywhere..

spring-data-neo4j/pom.xml
nmervaillie: I just applied the suggestions from Intellij. Looks a lot liek what's in the kotlin official docs.
I presume (maybe wrong) kotlin does not reimplement a full java compiler but delegates to javac.

src/main/asciidoc/reference/neo4j-repositories.adoc
nmervaillie: 1. No. Actually when the code in not compiled with -parameters flag, Spring Data falls back to byte code parsing to get argument names.

nmervaillie:
merged 2af6a45

PR #409, opened 2018-07-11
Datagraph - 1075 - Fix Page's element count calculation

meistermeier:
This PR will solve a problem with calculating the right count of total elements when using PageRequests.

### DIFF
--- a/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/support/SimpleNeo4jRepository.java
+++ b/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/support/SimpleNeo4jRepository.java
@@ -16,17 +16,16 @@ package org.springframework.data.neo4j.repository.support;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.List;
 import java.util.Optional;
 
 import org.neo4j.ogm.cypher.query.Pagination;
 import org.neo4j.ogm.session.Session;
 import org.springframework.data.domain.Page;
-import org.springframework.data.domain.PageImpl;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.neo4j.repository.Neo4jRepository;
 import org.springframework.data.neo4j.util.PagingAndSortingUtils;
+import org.springframework.data.repository.support.PageableExecutionUtils;
 import org.springframework.stereotype.Repository;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.util.Assert;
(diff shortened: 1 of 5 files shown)


spring-data-neo4j/src/test/java/org/springframework/data/neo4j/repository/support/SimpleNeo4jRepositoryTests.java
```diff
@@ -0,0 +1,91 @@
...
+import org.neo4j.ogm.cypher.query.SortOrder;
+import org.neo4j.ogm.session.Session;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.PageRequest;
+
+public class SimpleNeo4jRepositoryTests {
+
+	private final Session sessionMock = mock(Session.class);
+	private final PageRequest twoElementPageRequest = PageRequest.of(0, 2);
+	private final PageRequest oneElementPageRequest = PageRequest.of(0, 1);
```
michael-simons: I'd like to see a test in the middle of something, like 5 elements in total, requesting 2nd page with pagesize 2 as well.

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/support/SimpleNeo4jRepository.java
```diff
@@ -190,24 +189,9 @@ public void deleteAll() {
...
-	private Page<T> updatePage(Pageable pageable, List<T> results) {
-
-		int pageSize = pageable.getPageSize();
-		long pageOffset = pageable.getOffset();
-		long total = pageOffset + results.size() + (results.size() == pageSize ? pageSize : 0);
+		Pagination pagination = new Pagination(pageable.getPageNumber(), pageable.getPageSize());
+		Collection<T> data = session.loadAll(clazz, PagingAndSortingUtils.convert(pageable.getSort()), pagination, depth);
 
-		return new PageImpl<T>(results, pageable, total);
+		return PageableExecutionUtils.getPage(new ArrayList<>(data), pageable, () -> session.countEntitiesOfType(clazz));
```
michael-simons: To bad that `getPage` needs a list, not a collection. I just checked, we return a `Set` and so we are copying stuff around.

meistermeier: There will be no API change in SD commons in the near future. As far as we checked this `Collection` is already an `ArrayList` but I won't do a downcast here.

michael-simons: Fully understandable.

meistermeier:
Merged with commit cf3ae693e46b7107eb535040d08427b73d6f3045.

PR #411, opened 2018-07-20
DATAGRAPH-1094 - Neo4jRepositories don't work with multiple Neo4j instances.

michael-simons:
This changes the way the shared session and `Neo4jRepositoryConfigurationExtension` are registered as root beans:

1. the shared session is no longer called `SharedSessionCreator`. The `SharedSessionCreator` is only the factory creating new shared sessions, so that was misleading when reading the code. @meistermeier and I opted for an ok in this possible breaking change, as hardly anyone would have injected a `Session` and added a `@Qualifier("sharedSessionCreator")`
2. Add a qualifier to the session using the name of the `SessionFactory` from which the session is created. This is now consistent with the way Spring Data JPA works
3. if there are no root beans under the given name, nothing changes. If there's already a root bean with the given default bean, the `registerWithSourceAndGeneratedBeanName` helper is used.
4. The now uses the correct factory configured in `@EnableNeo4jRepositories`

Apart from the changed bean name, nothing changes when the user only has one `SessionFactory`. If he wants multiple, than we fully support this while still having an injectable session factory.

Formattings as suggested in DATAGRAPH-1093 are not applied yet and will follow. Also the test need some more polishing.

- [x] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [x] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATAGRAPH).
- [x] You submit test cases (unit or integration tests) that back your changes.

### DIFF
--- a/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/config/Neo4jMappingContextFactoryBean.java
+++ b/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/config/Neo4jMappingContextFactoryBean.java
@@ -21,6 +21,7 @@ import org.springframework.beans.factory.config.AbstractFactoryBean;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.data.neo4j.mapping.Neo4jMappingContext;
+import org.springframework.util.Assert;
 
 /**
  * {@link FactoryBean} to setup {@link Neo4jMappingContext} instances from Spring configuration.
@@ -37,10 +38,11 @@ public class Neo4jMappingContextFactoryBean extends AbstractFactoryBean<Neo4jMap
 	private ListableBeanFactory beanFactory;
 
 	/**
-	 * Uses the default session factory name {@link Neo4jRepositoryConfigurationExtension#DEFAULT_SESSION_FACTORY_BEAN_NAME}
-	 * for finding the session factory passed to the {@link Neo4jMappingContext}.
+	 * Uses the default session factory name
+	 * {@link Neo4jRepositoryConfigurationExtension#DEFAULT_SESSION_FACTORY_BEAN_NAME} for finding the session factory
+	 * passed to the {@link Neo4jMappingContext}.
 	 *
-	 * @deprecated
+	 * @deprecated since 5.1.1, use {@link #Neo4jMappingContextFactoryBean(String)} instead
 	 */
 	public Neo4jMappingContextFactoryBean() {
 		this(Neo4jRepositoryConfigurationExtension.DEFAULT_SESSION_FACTORY_BEAN_NAME);
@@ -49,10 +51,12 @@ public class Neo4jMappingContextFactoryBean extends AbstractFactoryBean<Neo4jMap
 	/**
 	 * Configures the mapping context with a named {@link SessionFactory}.
 	 *
-	 * @param sessionFactoryBeanName
+	 * @param sessionFactoryBeanName must not be {@literal null} or empty.
 	 * @since 5.1.0
 	 */
-	Neo4jMappingContextFactoryBean(String sessionFactoryBeanName) {
+	public Neo4jMappingContextFactoryBean(String sessionFactoryBeanName) {
+
+		Assert.hasText(sessionFactoryBeanName, "SessionFactoryBeanName must not be null nor empty!");
 		this.sessionFactoryBeanName = sessionFactoryBeanName;
 	}
 
--- a/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/config/Neo4jMappingContextFactoryBean.java
+++ b/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/config/Neo4jMappingContextFactoryBean.java
@@ -27,11 +27,34 @@ import org.springframework.data.neo4j.mapping.Neo4jMappingContext;
  *
  * @author Mark Angrish
  * @author Nicolas Mervaillie
+ * @author Michael J. Simons
  */
 public class Neo4jMappingContextFactoryBean extends AbstractFactoryBean<Neo4jMappingContext> implements
 		ApplicationContextAware {
 
+	private final String sessionFactoryBeanName;
+
 	private ListableBeanFactory beanFactory;
+
+	/**
+	 * Uses the default session factory name {@link Neo4jRepositoryConfigurationExtension#DEFAULT_SESSION_FACTORY_BEAN_NAME}
+	 * for finding the session factory passed to the {@link Neo4jMappingContext}.
+	 *
+	 * @deprecated
+	 */
+	public Neo4jMappingContextFactoryBean() {
+		this(Neo4jRepositoryConfigurationExtension.DEFAULT_SESSION_FACTORY_BEAN_NAME);
+	}
+
+	/**
+	 * Configures the mapping context with a named {@link SessionFactory}.
+	 *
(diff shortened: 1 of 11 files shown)


spring-data-neo4j/src/test/java/org/springframework/data/neo4j/repository/MultipleSessionFactorySupportTests.java
```diff
@@ -0,0 +1,225 @@
...
+	private long executeCountQuery(String countQuery, GraphDatabaseService onInstance) {
+
+		try (Transaction ignored = onInstance.beginTx()) {
+			return (long) onInstance.execute(countQuery).next().get("cnt");
+		}
+	}
+
+	private static DatabaseHolder createAndStartEmbeddedInstance() throws IOException {
+
+		final File dataStore1 = Files.createTempDirectory("neo4j.db").toFile();
```
mp911de: Why `final` at variable level?

spring-data-neo4j/src/test/java/org/springframework/data/neo4j/repository/config/Neo4jRepositoryConfigurationExtensionTests.java
```diff
@@ -44,35 +46,29 @@
  * @author Mark Angrish
  * @author Mark Paluch
  * @author Gerrit Meier
+ * @author Michael J. Simons
  */
 public class Neo4jRepositoryConfigurationExtensionTests {
```
mp911de: How about another test to verify bean registration with a generated name?

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/config/Neo4jMappingContextFactoryBean.java
```diff
@@ -27,12 +27,35 @@
...
+
 	private ListableBeanFactory beanFactory;
 
+	/**
+	 * Uses the default session factory name {@link Neo4jRepositoryConfigurationExtension#DEFAULT_SESSION_FACTORY_BEAN_NAME}
+	 * for finding the session factory passed to the {@link Neo4jMappingContext}.
+	 *
+	 * @deprecated
+	 */
+	public Neo4jMappingContextFactoryBean() {
```
mp911de: Missing `@Deprecated` on constructor level.

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/config/Neo4jMappingContextFactoryBean.java
```diff
@@ -27,12 +27,35 @@
...
 
+	private final String sessionFactoryBeanName;
+
 	private ListableBeanFactory beanFactory;
 
+	/**
+	 * Uses the default session factory name {@link Neo4jRepositoryConfigurationExtension#DEFAULT_SESSION_FACTORY_BEAN_NAME}
+	 * for finding the session factory passed to the {@link Neo4jMappingContext}.
+	 *
+	 * @deprecated
```
mp911de: Adding `@deprecated since <version>, use … instead` is quite helpful to users.

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/config/Neo4jMappingContextFactoryBean.java
```diff
@@ -27,12 +27,35 @@
...
+	}
+
+	/**
+	 * Configures the mapping context with a named {@link SessionFactory}.
+	 *
+	 * @param sessionFactoryBeanName
+	 * @since 5.1.0
+	 */
+	Neo4jMappingContextFactoryBean(String sessionFactoryBeanName) {
+		this.sessionFactoryBeanName = sessionFactoryBeanName;
```
mp911de: How about adding `Assert.notEmpty(…)` for `sessionFactoryBeanName`? Makes also sense to document the constraint in the javadoc. Mind also turning the constructor into a `public` one so the class can be instantiated from outside the package?

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/config/Neo4jRepositoryConfigurationExtension.java
```diff
@@ -168,35 +185,74 @@ public void registerBeansForRoot(BeanDefinitionRegistry registry, RepositoryConf
 
 		Object source = config.getSource();
 
-		registerIfNotAlreadyRegistered(createSharedSessionCreatorBeanDefinition(config), registry,
-				NEO4J_SHARED_SESSION_CREATOR_BEAN_NAME, source);
+		this.sessionBeanName = registerWithSuggestedNameOrGenerateNewName(createSharedSessionCreatorBeanDefinition(config),
```
mp911de: Bean registration should either use generated names or provided ones, see also for the mapping context. Mixing behavior can be confusing.

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/support/Neo4jRepositoryFactoryBean.java
```diff
@@ -66,18 +67,12 @@ public void afterPropertiesSet() {
...
-		return createRepositoryFactory(session);
-	}
-
-	/**
-	 * Returns a {@link RepositoryFactorySupport}.
-	 *
-	 * @param session
-	 * @return
-	 */
-	protected RepositoryFactorySupport createRepositoryFactory(Session session) {
```
mp911de: How about deprecating the method first? `Neo4jRepositoryFactoryBean` isn't the typical user-facing API, yet there might be someone that has customized `createRepositoryFactory(…)`.

spring-data-neo4j/src/test/java/org/springframework/data/neo4j/repository/MultipleSessionFactorySupportTests.java
```diff
@@ -0,0 +1,225 @@
...
+
+/**
+ * DATAGRAPH-1094
+ *
+ * @author Michael J. Simons
+ */
+@ContextConfiguration(classes = { MultipleSessionFactorySupportTests.BaseConfiguration.class,
+		MultipleSessionFactorySupportTests.FriendsConfiguration.class,
+		MultipleSessionFactorySupportTests.RestaurantsConfiguration.class })
+@RunWith(SpringJUnit4ClassRunner.class)
```
mp911de: How about `@RunWith(SpringRunner.class)`?

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/transaction/SharedSessionCreator.java
```diff
@@ -60,7 +61,6 @@
 	 * @return a shareable transactional Session proxy
 	 */
 	public static Session createSharedSession(SessionFactory sessionFactory) {
-
```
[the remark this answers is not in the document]
michael-simons: Ok, I'll try to mimic Jens' approach.

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/config/Neo4jMappingContextFactoryBean.java
michael-simons: I don't understand. It's there, so it should be missing?

mp911de:
That's merged and polished now.

PR #412, opened 2018-07-20
Issue/datagraph 1099

michael-simons:
We should not use deprecated things in our own example.


meistermeier:
Additional commit as a suggestion to widen the focus also on the `@Index` `primary` attribute.

PR #416, opened 2018-07-24
Support version property information

meistermeier:
Let Spring Data Rest pick up the right version property to provide its value as an Etag.


michael-simons:
Closed with 2748aedd3ae85bc6cfab17cb1ba51dcbf5174941.

PR #418, opened 2018-08-08
DATAGRAPH-1109 - Add support for persistence constructors for @QueryResults

michael-simons:
We support persistence constructors for classes annotated as `@NodeEntity` but not for projections based on DTOS annotated with `@QueryResult`.

This PR provides persistence constructors for those projections as well, reducing the amount of different programming models a user has to deal with.

The PR is centered around the idea to make `OgmEntityInstantiatorAdapter` a public api as `Neo4jOgmEntityInstantiatorAdapter` in the conversion package much as similar concepts in the Mongo store. It also drops the idea of `OgmReflectionEntityInstantiator` being solely an indirection against OGMs 3.1 `ReflectionEntityInstantiator`. The guard whether the later is available is still applied (in `CustomResultConverter`).

`OgmReflectionEntityInstantiator` becomes `QueryResultInstantiator` as dedicated means of instantiating DTOs from a custom result. It either uses the `ReflectionEntityInstantiator` as before or the new public api `Neo4jOgmEntityInstantiatorAdapter` if applicable.

It is applicable if it get hold of the `Neo4jMappingContext`. 

To pass this around, the `Neo4jRepositoryFactoryBean` and `Neo4jRepositoryFactory` have been extended to transport it up to the `GraphQueryMethod`, again, much as the Mongo store does.

From there on it's accessible in all query related scenarios.


meistermeier:
Merged with commit aa9ba9e

PR #424, opened 2018-09-05
DATAGRAPH-1120 - Cleanup derived query infrastructure.

michael-simons:
This PR rewrites the infrastructure for computing derived queries.

It doesn't change the base logic, but makes the creation of derived query understandable so that we or others can bring in new features or bug fixes without having a rough start.

The `DerivedGraphRepositoryQuery` still keeps a query template without resolved parameters around, so that the work going through all parts is done only once.

The changes make clear that the implementation of `AbstractQueryCreator` now named `FilterBuildersQueryCreator` creates a query based on filters, not on cypher. The `FilterBuildersQuery` makes clear that its not executable (Note: I am still thinking naming it `XXXTemplate`). 

The PR does change several public classes but not classes that are officially part of an API. People wo dug that deep that they would have supplied their on `AbstractQueryCreator` based on the previous implementation will have to do some work, but for a developer using SDN 5.1 nothing changes.


mp911de:
It's quite late in the release cycle to touch the API. However, I suggest to align naming (Using rather `PartTreeNeo4jQuery` instead of `DerivedGraphRepositoryQuery` and such, see MongoDB and JPA modules for further details) and structuring of packages entirely. Typically, Query Creator API is not used in the application code and therefore it should not break applications.

michael-simons:
Thanks a lot Mark. Good catch. It's one of our main goals to align SDN as much was the other projects as possible to be able to keep your pace some point in the future.

We have the same opinion about the query creator API being private API, but I'm totally open about your judgement here. If we decide to include that change in Lovelace, that's on us anyway.

mp911de:
Looks better now.

meistermeier:
Merged with 025b877d96e3467359e92edae47ac7845bbb5eb1 and fe36704ab54d0befdeaf4f87b70495cd4660e14f

PR #425, opened 2018-09-10
DATAGRAPH-1122  - Prevent eager initialization of beans for repository support.

michael-simons:
I'm just using the newly introduced methods in Commons 2.1 here (and thus remove calls do deprecated methods).


michael-simons:
(Only Lovelace)

meistermeier:
Merged with 94115073

PR #429, opened 2018-09-19
DATAGRAPH-1126 Fix application startup with multiple ConversionServices

michael-simons:
4 commits here. The first 3 are required and should be picked/squash. They fix the issue at hand.

Have a sharp look at number 4: It will break things, i.e. if people brought in SDN without actually using it but calling the constructor of MetaDataDrivenConversionService and relying on the fact that it does all kinds of stuff or requiring MetaDataDrivenConversionService to implement the callback interface.

However, this stuff is new anyway in 5.1, so better break it now than in 5.2.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j-parent</artifactId>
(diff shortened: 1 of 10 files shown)


michael-simons:
Build was/is green, I double checked it in a local setup agains Spring Boot 2.1.0.M3 with Spring Data Rest included as well.

PR #433, opened 2018-10-15
DATAGRAPH-1135 Fix resolution of implicitly named parameters on Repository queries.

michael-simons:
This checks wether a Parameter is actually a named parameter (not special and having a name determinable by some means) and if so, adds it to map of parameters.

We still add the parameter based on its index, otherwise we would break existing behavior (parameters can be used mixed: By index and name at the same time).

There is another PR incoming for the 5.0.x series. It keeps the logic, but skips the polishing due to other changes.


meistermeier:
merged with bb1205b2

PR #435, opened 2018-10-15
DATAGRAPH-1141 - Make SDN 5.1.1 compatible with Spring 5.0.x again.

michael-simons:
We should also have a look at having our own integration build / test matrix here.


meistermeier:
closed with e874238 and 702a52a

PR #436, opened 2018-10-23
DATAGRAPH-1144 Recognize not only internal ids as id properties.

michael-simons:
From the main commit:

Fix the determination of Spring Data Id Properties:

Before the fix, Neo4jPersistentProperty only considered the „identityField“ from OGMs ClassInfo as id property. This identifyField maps only to Neo4j internal, native ids.

As soon as a user annotates a non-long field with Id Spring Data cannot determine a required id property for the entity anymore.
While the entity can still be found by the „external“ id due to the fact that OGM keeps a cache mapping from external to internal / native, modules depending on knowing the property attribute, break.

The fix introduces a property type, which can be regular, internal id or id property.

Id properties are preferred over internal id properties based on the identityField as SDN id propertiers. This has the byproduckt that we now can have an „external“ id (the property mapped with @Id) and a „Long id“ field in an entity so that a developer can refer to both.

Also take note, that I'll provide a fix for the fact that `org.springframework.data.neo4j.repository.Neo4jRepositoryTests` works only exactly in the transactional setup and the deleteAll will fail in a different transaction with OGM later this wake.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j-parent</artifactId>
-	<version>5.2.0.BUILD-SNAPSHOT</version>
+	<version>5.2.0.DATAGRAPH-1144-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data Neo4j</name>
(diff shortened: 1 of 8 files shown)


meistermeier:
Merged with f9ea13a9e3486e6ecf4e9bc966883a3c9a37c21d (master) and 6ce311338b0b96000515e49557aaa4ad4dd610b7 (5.1)

PR #439, opened 2018-11-07
DATAGRAPH-1149 - Introduce MetaDataProvider to avoid unnecessary transactions.

michael-simons:
The newly introduced MetaDataProvider is used to retrieve the OGM MetaData from the session factory as the „source of truth.“ Getting them from the session itself is hacky and can lead to the creation of a huge amount of unnessary transactions.

As SDN works only on the session object and keeps the session factory hidden, we faciliate the shared session creator to add this interface to the session and delegate back to the session factory. If the proxy isn’t used, we’ll try to lookup the meta data on the concrete Neo4jSession.

Two test that mock the session and don’t use the proxy had to be fixed to include the additional interface.

### DIFF
--- a/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/mapping/MetaDataProvider.java
+++ b/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/mapping/MetaDataProvider.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c)  [2011-2018] "Pivotal Software, Inc." / "Neo Technology"
+ *
+ * This product is licensed to you under the Apache License, Version 2.0 (the "License").
+ * You may not use this product except in compliance with the License.
+ *
+ * This product may include a number of subcomponents with
+ * separate copyright notices and license terms. Your use of the source
+ * code for these subcomponents is subject to the terms and
+ * conditions of the subcomponent's license, as noted in the LICENSE file.
+ *
+ */
+package org.springframework.data.neo4j.mapping;
+
+import org.neo4j.ogm.metadata.MetaData;
+
+/**
+ * The only meta-data provider currently is a Neo4j-OGM {@link org.neo4j.ogm.session.SessionFactory} by proxy. The
+ * meta-data is needed during execution of graph-repository-queries to lookup mappings etc.
+ *
+ * @author Michael J. Simons
+ */
+@FunctionalInterface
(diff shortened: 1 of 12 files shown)


spring-data-neo4j/src/main/java/org/springframework/data/neo4j/mapping/MetaDataProvider.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package org.springframework.data.neo4j.mapping;
+
+import org.neo4j.ogm.metadata.MetaData;
+
+/**
+ * The only meta-data provider currently is a Neo4j-OGM {@link org.neo4j.ogm.session.SessionFactory} by proxy. The
+ * meta-data is needed during execution of graph-repository-queries to lookup mappings etc.
+ *
+ * @author Michael J. Simons
```
mp911de:
```suggestion
 * @author Michael J. Simons
 * @since …
```

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/query/AbstractGraphRepositoryQuery.java
```diff
@@ -28,12 +29,14 @@
...
 public abstract class AbstractGraphRepositoryQuery implements RepositoryQuery {
 
-	private final GraphQueryMethod method;
-	private final Session session;
+	protected final GraphQueryMethod queryMethod;
+	protected final MetaData metaData;
+	protected final Session session;
 
-	protected AbstractGraphRepositoryQuery(GraphQueryMethod method, Session session) {
+	protected AbstractGraphRepositoryQuery(GraphQueryMethod queryMethod, MetaData metaData, Session session) {
```
mp911de: You might want to keep polishing in a separate commit making the actual change more obvious.

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/mapping/MetaDataProvider.java
michael-simons: I realized to late that you used the suggested change feature. Next time.

michael-simons:
As 9d10803e744036130d79e1431a4b3b3b5dff9df0 and 3ca298ad2d705b78b78829730bda22390cb7f96e on master now.

PR #440, opened 2018-11-07
DATAGRAPH-1150 - Align Kotlin Version with Spring Data Commons.

michael-simons:
I cannot stand the compiler warnings. Also I have a bad feeling mixing those versions, especially in hindsight of the work we're planing regarding immutability.

Should be cherry picked into master as well as into 5.1.x. Thanks.


meistermeier:
Merged with 7d11b5a2bce53981e9c78877217a6bb8b9bb0ecf

PR #448, opened 2019-01-08
DATAGRAPH-1174 - Fix class loading leak with custom QueryResults. (Lovelace)

michael-simons:
The changes we discussed. Feel free to squash them as needed.

8e00877 is optional, but I think it's worth it.


michael-simons:
Done too.

PR #452, opened 2019-01-30
Document limitations of auditing and read-only transactions

meistermeier:
This should also go into the 5.1.x branch.

### DIFF
--- a/src/main/asciidoc/reference/auditing.adoc
+++ b/src/main/asciidoc/reference/auditing.adoc
@@ -18,4 +18,10 @@ import org.springframework.data.neo4j.annotation.EnableNeo4jAuditing;
 public class Neo4jConfiguration {}
 ----
 
+[NOTE]
+If you are working with custom ids by setting them directly or using your own `IdStrategy`, the auditing support is limited.
+In this case it is not possible for Spring Data Neo4j to figure out if an entity is new or already exists in the database,
+because it has an id field set.
+As a consequence the fields annotated with `@CreatedBy` and `@CreatedDate` will never get filled.
+
 Please refer to the <<auditing,auditing>> section of the shared Spring Data reference for further details.
--- a/src/main/asciidoc/reference/auditing.adoc
+++ b/src/main/asciidoc/reference/auditing.adoc
@@ -20,8 +20,7 @@ public class Neo4jConfiguration {}
 
 [NOTE]
 If you are working with custom ids by setting them directly or using your own `IdStrategy`, the auditing support is limited.
-In this case it is not possible for Spring Data Neo4j to figure out if an entity is new or already exists in the database,
-because it has an id field set.
+It it is not possible for Spring Data Neo4j to figure out if an entity is new or already exists in the database.
 As a consequence the fields annotated with `@CreatedBy` and `@CreatedDate` will never get filled.
 
 Please refer to the <<auditing,auditing>> section of the shared Spring Data reference for further details.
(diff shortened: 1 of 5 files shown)


src/main/asciidoc/reference/auditing.adoc
```diff
@@ -18,4 +18,10 @@ import org.springframework.data.neo4j.annotation.EnableNeo4jAuditing;
 public class Neo4jConfiguration {}
 ----
 
-Please refer to the <<auditing,auditing>> section of the shared Spring Data reference for further details.
\ No newline at end of file
+[NOTE]
+If you are working with custom ids by setting them directly or using your own `IdStrategy`, the auditing support is limited.
+In this case it is not possible for Spring Data Neo4j to figure out if an entity is new or already exists in the database,
```
michael-simons:
```suggestion
It is not possible for Spring Data Neo4j to figure out if an entity is new or already exists in the database if the id field is set through a custom strategy.
```

src/main/asciidoc/reference/auditing.adoc
```diff
@@ -18,4 +18,10 @@ import org.springframework.data.neo4j.annotation.EnableNeo4jAuditing;
...
+because it has an id field set.
```
michael-simons:
```suggestion
```

src/main/asciidoc/reference/neo4j-repositories.adoc
```diff
@@ -482,8 +482,12 @@ Below are some excellent resources:
 
 You can start a read only transaction by marking a class or method with `@Transactional(readOnly=true)`.
 
+[NOTE]
+A custom cypher provided by using the `@Query` annotation will automatically force a read/write transaction.
+This is a known limitation because Neo4j-OGM ignores the `readOnly` flag if custom cypher is provided.
+
 [CAUTION]
-Note that if you open a read only transaction from a service method and then call a mutating method that is marked as read/write your transaction semantics will always be defined by the outermost transaction. 
+Note that if you open a read only transaction from a service method and then call a mutating method that is marked as read/write your transaction semantics will always be defined by the outermost transaction.
```
michael-simons: I don't understand this.

I'll get a read transaction, right? Should we link my article about validating transactions? Or the docs? Or the snippet how to do this.

src/main/asciidoc/reference/auditing.adoc
meistermeier: Custom strategy, as mentioned in the sentence before, is just one of the two possibilities to provide an id besides the internal (native id). If we would only focus on custom strategy (the term) here it might sound more confusing. But yes, also my expression could lead to the same assumption the other way round.

src/main/asciidoc/reference/neo4j-repositories.adoc
meistermeier: Could you please suggest something for this? I think it should then get out of the "caution box". The commit touches this because of the trailing whitespace but there was no intention to change anything.

michael-simons: Something like "Be careful when nesting transactions. Spring's `@Transactional` propagates by default outer transactions and doesn't validate whether they are compatible with inner. See [link to Spring docs.]"

meistermeier:
Closed with 4162128fe187b5243dd902513b729e765ca6c1a4 and ed6b2d7f602f013045027a113ae73330d451571b for master.
c46d0ee8f91c0cd2d97ba64f1497cb6321137d67 and 68db9f905e1895df660f1e20764944ab2222fb85 for 5.1.x

PR #453, opened 2019-02-01
Allow concatenation of ignoreCase modifier keyword.

meistermeier:
Add support for using `containing` with the `ignoreCase` keyword in derived finder methods. 

This PR can only get merged in master after upgrading to Neo4j-OGM 3.2.0-alpha04.


meistermeier:
Merged with 84a9e390aa60c757349d7a7ef256e07c6201e1ae.

PR #457, opened 2019-02-15
Support exists queries

meistermeier:
An exists(By) keyword support for derived query methods, that does evaluate the non-zero positive result of a filter query.
For custom query checks provide an `exists` flag in the `@Query` annotation to mark it as eligible for boolean returns. If the statement produces any results, the returned value will be true.

Some thoughts for further improvement/alignment with previous behaviour:
We could think of also introducing a wrapping annotation like Spring Data Cassandra does: https://github.com/spring-projects/spring-data-cassandra/blob/e6b5c5e7a2af5c020526e6d998d5950b74163937/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/ExistsQuery.java#L37

Additionally this PR breaks with the previous convention in the count query of providing a String value but only flags the `value` of the `@Query` annotation with `exists = true/false`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j-parent</artifactId>
-	<version>5.2.0.BUILD-SNAPSHOT</version>
+	<version>5.2.0.DATAGRAPH-1197-SNAPSHOT</version>
 	<packaging>pom</packaging>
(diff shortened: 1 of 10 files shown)


meistermeier:
Merged with 56ac08d3dc6e2d3975cdaf792fb9a97d8541eacb and 55d6c925227fff257dfb1e908139894ab9c63964

PR #458, opened 2019-02-19
Provide @ExistsQuery for convenience

meistermeier:
Add an `@ExistsQuery` annotation that wraps a `@Query(exists =true)` to have a more convenient way to define custom queries that should get used for boolean projection.


meistermeier:
Merged with d53ec47b369ac8cb0bfeb10b19c56a4279148971

PR #459, opened 2019-03-01
Add support for Range type in derived queries

meistermeier:
Additional to the two parameter (lower and upper bound) that uses a fixed inclusive pattern it is now possible to provide a `Range` to express inclusive or exclusive bounds explicitly.

### DIFF
--- a/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/query/filter/BetweenComparisonBuilder.java
+++ b/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/query/filter/BetweenComparisonBuilder.java
@@ -15,6 +15,8 @@
  */
 package org.springframework.data.neo4j.repository.query.filter;
 
+import static org.neo4j.ogm.cypher.ComparisonOperator.*;
+
 import java.util.Arrays;
 import java.util.List;
 import java.util.Stack;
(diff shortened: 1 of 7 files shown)


spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/query/filter/BetweenComparisonBuilder.java
```diff
@@ -39,22 +42,71 @@
...
+
+		Filter upperBoundFilter = createUpperBoundFilter(upperBoundValue, inclusiveUpperBound);
+		setNestedAttributes(part, upperBoundFilter);
+
+		return Arrays.asList(lowerBoundFilter, upperBoundFilter);
+	}
+
+	private Filter createLowerBoundFilter(Object value, boolean inclusive) {
+		Filter filter = createBoundFilter(Bound.LOWER, value, inclusive);
+		filter.setBooleanOperator(booleanOperator);
```
michael-simons: Can we pass on the operator on both of those (lower and upper)? 
The "and" is used for the boolean operator between previous and this or this and next?

meistermeier: The "and" is for "greater/equals than lower bound AND lower/equals than upper bound". The `booleanOperator` (And/Or) is for the previous ones like `findByNameEqualsAnd/OrSomethingBetween`.

meistermeier:
Merged with 5de95cbde11fe1707e88d73f490650d88b3b6a04

PR #461, opened 2019-03-07
DATAGRAPH-1171 - Provide native Point to Spring Data Commons Point converter.

michael-simons:
Please have a detailed look at 

`org.springframework.data.neo4j.mapping.Neo4jMappingContext.NativePointConverter` and it's comment.

`org.springframework.data.neo4j.repository.query.filter.DistanceComparisonBuilder#createFilterForSpringPoint`
This can be done in OGM as well, but I think it's a pure SDN need…

Apart from that, I fixed the usage of distance in meter and added tests for that.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j-parent</artifactId>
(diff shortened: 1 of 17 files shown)


michael-simons:
Requires a new release of OGM with https://github.com/neo4j/neo4j-ogm/pull/616 included.

PR #475, opened 2019-03-21
Fix treatment of collection-like constructor parameters.

michael-simons:
See reasoning here

https://jira.spring.io/browse/DATAGRAPH-1210

Should be applied to both 5.1 and 5.2.

FYI @rlippolis.

### DIFF
--- a/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/conversion/Neo4jOgmEntityInstantiatorAdapter.java
+++ b/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/conversion/Neo4jOgmEntityInstantiatorAdapter.java
@@ -90,7 +90,7 @@ public class Neo4jOgmEntityInstantiatorAdapter implements org.neo4j.ogm.session.
 			}
 		}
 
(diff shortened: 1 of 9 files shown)


spring-data-neo4j/src/main/java/org/springframework/data/neo4j/conversion/Neo4jOgmEntityInstantiatorAdapter.java
```diff
@@ -77,12 +81,35 @@ public Neo4jOgmEntityInstantiatorAdapter(MappingContext<Neo4jPersistentEntity<?>
...
+
+			Object value = extractParameterValue(parameter);
 			if (value == null || conversionService == null) {
 				return value;
 			} else {
 				return conversionService.convert(value, parameter.getType().getType());
 			}
 		}
+
+		Object extractParameterValue(PreferredConstructor.Parameter parameter) {
```
meistermeier: Can be private, or?

michael-simons:
Cherry-picked in master and 5.1

PR #499, opened 2019-12-06
DATAGRAPH-1274 - Don't rely on Neo4j-OGM failing with NPE in tests.

michael-simons:
In theory OGM should have returned null or empty collections when it's queried by an unmapped class. Until 3.2.4 however it used to throw exceptions in some cases. We fixed this, but now we cannot test for it anymore (which was a stupid idea to begin with).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j-parent</artifactId>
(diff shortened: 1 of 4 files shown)


michael-simons:
Fixed in master as a529a0834c60a153fcf387d027e2a94e15429973 and as 0641830589a5ff7976811f60532ca821073c33c6 in 5.2.x.

PR #500, opened 2020-01-06
DATAGRAPH-1276 - Explicitly Pass READ_ONLY mode to Session.query.

michael-simons:
This change makes sure we pass the correct query mode even when people workaround the repositories and use the injected session directly.


michael-simons:
Merged as

* 5.3 (Master): af886623d907d7fa9fd1e702c003a45e07c9ce33
* 5.2: 3b7bf60e284a2204b3f97554c9cb53303e7a8d20
* 5.1: 3a217e4d4b34cc2de0fce5507953b96c15b05536

PR #506, opened 2020-02-13
DATAGRAPH-1290 - Add case insensitive order support for custom queries

ihordz:
- [x] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [x] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATAGRAPH).
- [x] You submit test cases (unit or integration tests) that back your changes.
- [x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/query/Query.java
+++ b/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/query/Query.java
@@ -129,10 +129,20 @@ public class Query {
 	}
 
 	private String getSortOrder(Sort sort) {
-
 		return sort.stream()
-				.map(order -> order.getProperty() + " " + order.getDirection())
+				.map(this::getPropertySortOrder)
 				.collect(Collectors.joining(", "));
+	}
+
+	private String getPropertySortOrder(Sort.Order order) {
+		StringBuilder sb = new StringBuilder();
+		if(order.isIgnoreCase()){
+			sb.append("toLower(").append(order.getProperty()).append(")");
+		} else {
+			sb.append(order.getProperty());
+		}
+		sb.append(" ").append(order.getDirection());
+		return sb.toString();
 	}
 
 	private String formatBaseQuery(String cypherQuery) {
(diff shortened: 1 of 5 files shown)


spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/query/Query.java
```diff
@@ -129,12 +129,22 @@ private String addSorting(String baseQuery, Sort sort) {
 	}
 
 	private String getSortOrder(Sort sort) {
-
 		return sort.stream()
-				.map(order -> order.getProperty() + " " + order.getDirection())
```
michael-simons: Very nice change already for that!

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/query/Query.java
```diff
@@ -129,12 +129,22 @@ private String addSorting(String baseQuery, Sort sort) {
...
+				.map(this::getPropertySortOrder)
 				.collect(Collectors.joining(", "));
 	}
 
+	private String getPropertySortOrder(Sort.Order order) {
```
michael-simons:
```suggestion
	private static String getPropertySortOrder(Sort.Order order) {
```

ihordz: done

spring-data-neo4j/src/main/java/org/springframework/data/neo4j/repository/query/Query.java
ihordz: thanks

ihordz:
@michael-simons 
I've pushed requested changes.
Thank you for review.

michael-simons:
This is now on master as 1ec70c426fa1489e2850a2f5995c78c5d37d5dd4 and on 5.2.x as 0bcf9f22d44e60e665c674f835e2c4cc6fd6226a. On 5.2.x I rewrote the tests to use Junit 4 asserts, I didn't want to have mixed styles in the class, but that's an oversight on my side.

Your PRs are adding nice value, thank you, @ihordz .

PR #519, opened 2020-08-09
DATAGRAPH-1356 - Add support for java.net.URL attributes.

michael-simons:


dcrissman:
Thank you @michael-Simmons for polishing my PR..

michael-simons:
There was not too much to polish :) Thank you for the contributions! 

I'm gonna bring in `java.net.URI` for completeness as well. 

A tip regarding `java.net.URL`: If you use the class extensively and are depending on it's equals method, that can be quite slow and even problematic in reactive database access, as `equals` is a blocking method that actually resolves hosts names, see for reference: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URL.html#equals(java.lang.Object)

michael-simons:
In case you wondering why I created a new PR: To convince CI to build it for me.

PR #529, opened 2020-09-10
Refactor relationships with properties

meistermeier:
New API that brings the `Map<Entity, RelationshipProperty>` construct into a `Collection<RelationshipProperty>` format.
The relationship property holds the target entity defined with `@TargetNode`.

There is no support for bi-directional relationships with properties right now. This will hopefully be done in future change.


michael-simons:
Merged as 4e2a18ba1e7460d729a6a6b5306e6e49e1069ae9

PR #537, opened 2020-09-26
 DATAGRAPH-1389 - Adopt to changes in Spring Data Commons

mp911de:
Check for collection-like return types in `ReactiveNeo4jQueryMethod` to determine whether the defined method return type is multi-valued. Also, make sure to call the overridden methods instead of defaulting to the superclass implementation.

Enable reusable testcontainers to reduce build time from ~ 8 minutes to ~ 3 minutes. Use a manual container lifecycle to avoid closing the container by the test so Ryuk takes care of the shutdown when the reuse feature isn't enabled via `~/.testcontainers.properties` (`testcontainers.reuse.enable=true`).

- DATAGRAPH-1389 - Prepare issue branch.
- DATAGRAPH-1389 - Adopt to changes in Spring Data Commons.
- DATAGRAPH-1389 - Polishing.

### DIFF
--- a/src/main/java/org/springframework/data/neo4j/repository/query/ReactiveNeo4jQueryMethod.java
+++ b/src/main/java/org/springframework/data/neo4j/repository/query/ReactiveNeo4jQueryMethod.java
@@ -25,6 +25,7 @@ import org.springframework.data.projection.ProjectionFactory;
 import org.springframework.data.repository.core.RepositoryMetadata;
 import org.springframework.data.repository.util.ReactiveWrappers;
 import org.springframework.data.util.ClassTypeInformation;
+import org.springframework.data.util.Lazy;
 import org.springframework.data.util.TypeInformation;
 import org.springframework.util.ClassUtils;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/neo4j/repository/query/ReactiveNeo4jQueryMethod.java
```diff
@@ -72,15 +76,35 @@
...
+	public boolean isCollectionQuery() {
+		return isCollectionQuery.get();
 	}
 
 	/**
-	 * Will always return true because a reactive result will always be a stream query.
+	 * Will always return false because it's not a {@link Stream}.
 	 *
-	 * @return always true
+	 * @return always false
```
michael-simons, line 96: 🤔

michael-simons (changes requested):
Comment and method say different.

src/main/java/org/springframework/data/neo4j/repository/query/ReactiveNeo4jQueryMethod.java
mp911de: Done.

michael-simons (approved):

michael-simons:
Thanks a lot, Mark! Also for cutting the build time in half. Merged as 31826386212483f94d61d7c3621e676e2c4b89c1.

PR #538, opened 2020-09-28
Provide attribute specific conversions and some legacy converters.

michael-simons:
This is dependent on #536 and contains to separate commits with both detailed messages what happened.


michael-simons:
Sorry, picked wrong branch.

PR #540, opened 2020-09-30
Temporarily save objects in creation

meistermeier:
This avoids the problem that an object in creation can get created multiple
times if there is a cycle or similar in the graph.


src/main/java/org/springframework/data/neo4j/core/mapping/DefaultNeo4jEntityConverter.java
```diff
@@ -476,40 +492,36 @@ private static Value extractValueOf(Neo4jPersistentProperty property, MapAccesso
...
+			}
 			try {
 
 				read.lock();
 
 				Object knownEntity = internalIdStore.get(internalId);
 
 				if (knownEntity != null) {
-					return knownEntity;
+					return Optional.of(knownEntity);
```
michael-simons, line 518: Can we do without the optionals here?

michael-simons (changes requested):

src/main/java/org/springframework/data/neo4j/core/mapping/DefaultNeo4jEntityConverter.java
meistermeier: For references: Exactly this change was done in the follow up PR #541

meistermeier:
Merged by 17a9235

PR #541, opened 2020-09-30
Optimize loading of cyclic data

meistermeier:
Currently the data for a model that forms a cycle containing
same node labels will be cut off on a certain level.
This commit changes the query creation and mapping to make use
of the path functionality in Neo4j.

Submitted for now as a draft PR because it depends on an unreleased version of the Neo4j CypherDSL.
Also it needs a little bit (a lot) of refactoring but should give a first impression of the change.


michael-simons (approved):

meistermeier:
Closed by 930f8fc

PR #542, opened 2020-10-02
DATAGRAPH-1393 - Add support for OGM's CompositeAttributeConverter.

michael-simons:
Like OGM, but better.

### DIFF
--- a/src/main/java/org/springframework/data/neo4j/core/mapping/DefaultNeo4jPersistentProperty.java
+++ b/src/main/java/org/springframework/data/neo4j/core/mapping/DefaultNeo4jPersistentProperty.java
@@ -72,7 +72,7 @@ final class DefaultNeo4jPersistentProperty extends AnnotationBasedPersistentProp
 
 			Class<?> targetType = getActualType();
 			return !(simpleTypeHolder.isSimpleType(targetType) || this.mappingContext.hasCustomWriteTarget(targetType)
-					|| isAnnotationPresent(TargetNode.class) || isAnnotationPresent(CompositeProperty.class));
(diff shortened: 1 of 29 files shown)


src/main/java/org/springframework/data/neo4j/core/mapping/DefaultNeo4jPersistentProperty.java
```diff
@@ -182,7 +183,7 @@ public boolean isAssociation() {
 
 	@Override
 	public boolean isEntity() {
-		return super.isEntity() && isAssociation() || (super.isEntity() && isEntityInRelationshipWithProperties());
+		return super.isEntity() && isAssociation() || (super.isEntity() && isEntityInRelationshipWithProperties() && !isAnnotationPresent(CompositeProperty.class));
```
meistermeier, line 186:
```suggestion
		return super.isEntity() && isAssociation() || (super.isEntity() && isEntityInRelationshipWithProperties() && !isComposite());
```

meistermeier: I prefer a central access to the annotation more than having the same (inversed) logic written again.

meistermeier (approved):
LGTM. As always just nitpicking.

src/main/java/org/springframework/data/neo4j/core/mapping/DefaultNeo4jPersistentProperty.java
michael-simons: Me too. Thanks.

michael-simons:
Merged as 7f75f13ef.

PR #544, opened 2020-10-06
DATAGRAPH-1387 - Move content of repository.core back to .support.

michael-simons:
See Slack discussion. The original reasoning behind this that everything - like in other parts of this and the other SD* frameworks - `support` packages contains often times classes that are useful to a user on their own (i.e. as base classes for config etc.) in contrast to providing framework extension points (which the factory beans are to me). 

Anyway, if that causes hassle downstream, I'm happy to fix it. Please merge as you see please.


mp911de (approved):

meistermeier:
Closed by 5c80380

PR #545, opened 2020-10-06
Support for TargetNode in constructor

meistermeier:
There was a limitation in the relationship properties mapping when it came to mapping the related entity as a constructor parameter in the `@RelationshipProperties` class.


michael-simons:
Merged as a9a59ffaf8c21e3f044b76eb66920e41c566d7fc

PR #546, opened 2020-10-09
DATAGRAPH-1400 - Use the correct domain type before creating DTO projections.

michael-simons:
This change makes sure we use the correct domain type before we create DTO based projections. Without taking the underlying domain type into consideration, annotated attributes, explicit relationship types etc. won't be taken into consideration when generating the Cypher query.

The DTO is than generated in two steps: First the domain type is instantiated and populated with the set of properties defined by the DTO and than the DTO is created from it. The creation is not part of the entity converter but a later step. In a future version of SDN this might change.

In addition to this changes, we allow additional attributes in a DTO projection. For them to be hydrated, custom queries must be supplied, containing the corresponding attributes.

Included with the commit is documentation of the defined behavior as well as a couple of tests to make sure the behavior of Spring Data Commons with regard to what is considered to be a projection or not keeps the way it is now.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j</artifactId>
-	<version>6.0.0-SNAPSHOT</version>
+	<version>6.0.0-DATAGRAPH-1400-SNAPSHOT</version>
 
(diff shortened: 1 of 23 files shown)


michael-simons:
Merged as b85a5beeb6532ed0f75f50216889a0594d650341.

PR #551, opened 2020-11-09
Support strict mode

meistermeier:
Introduce strict entity scanning.

I had to move integration tests into parallel packages to avoid converter registration in every integration test.
Removed one test because there is no reason to keep it as a separated case anymore IMHO.


src/test/java/org/springframework/data/neo4j/core/mapping/Neo4jMappingContextTest.java
```diff
@@ -184,6 +185,7 @@ public Object convert(Object source, TypeDescriptor sourceType, TypeDescriptor t
 	}
 
 	@Test
+	@Disabled("There is no implicit registration of entities anymore")
```
michael-simons, line 188: Sollten wir das denn nicht für den test enablen können? `setStrict(false)`?

michael-simons (changes requested):

src/test/java/org/springframework/data/neo4j/core/mapping/Neo4jMappingContextTest.java
meistermeier: Sollten wir :) (done)

PR #552, opened 2020-11-11
Allow aggregating into 1 single collection.

michael-simons:
We have everything in place to call our entity reader several times. The biggest change here is to pass in the map accessor from the outside, not the record, which I should have done when I changed the entity reader to work on the map accessor in the first place.

This allows the flight scenario and a couple of other things to work.

I would not go as far to support maps as we did in OGM as return type (that can be solved by the user in a custom streamable).

This PR has a couple of todos that needs to be implemented.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j</artifactId>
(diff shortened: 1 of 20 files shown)


meistermeier (approved):

michael-simons:
Merge in master and 6.0.x.

PR #555, opened 2020-11-16
Fix mapping of relationships.

meistermeier:


michael-simons:
Merged in master and 6.0.x

PR #557, opened 2020-11-19
Improve path-based query generation.

meistermeier:
Now the query generator 
a) tries to respect the directions of the relationships and make them explicit if they are homogenous
b) defines the path at long it does not enter the schematic self-reference / cycle, after this it falls back to `[:ALL|RELATIONSHIP|TYPES*0...]`


meistermeier:
Closed with c1cb92f848e93192af19490541e78132a87f732c (master) and b0a028546a2503725ff0278bae8b3b1ff58e1344 (6.0)

PR #558, opened 2020-11-19
Fix query generation for same label derived finders

meistermeier:
This fixes the problem for a scenario where `(:Person)-[:KNOWS]->(:Person)` and a derived finder method like `Person findByFriendName(String name)`.
The resulting query creation would not put the `name` onto the related person but on the root entity.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j</artifactId>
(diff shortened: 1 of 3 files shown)


michael-simons (approved):
Seems reasonable. What happens with `findByFriendsNameFriendsName("Pet3")` though?

meistermeier:
Closed with f21e7c9a5c9207bd5b323a8be636206f10e18025 (master) and 0dd0359912be90e2a3e4aa15eb78b816268db58a (6.0)

PR #559, opened 2020-11-19
Fix naming of relationships in query

meistermeier:
The queries for relationships with properties created a repetitive `__relationship__` definition.
I thought this will be nested and only visible in its scope. Turns out the inner scope gets the outer scope added and the database did not return any result.


michael-simons (approved):

meistermeier:
Closed with 89f3f08769eb3de29e41605983481514afdd02e1 (master) and 5ff542e6a275cc27863912f146eb0335ac2db635 (6.0)

PR #562, opened 2020-11-25
Avoid redundant statements for relationships.

meistermeier:
If bidirectional relationships are defined and set on both ends,
SDN created multiple create and delete statements.
This should happen only once for the very same relationship.


michael-simons:
Merged as 06360923201715889b6bcb4b4346190a45769923 on master, as 606f9f87b49da34dceb076bf6e816982eca00e55 on 6.0.x 👍

PR #563, opened 2020-12-08
Register Neo4jEvaluationContextExtension only once.

meistermeier:
This is now wrapped in a `registerIfNotAlreadyRegistered` block to ensure that it will get at most once registered.

@mp911de could you please give me a sanity check on this? I think it's the right way to keep do this but would be good to know if I have overseen something even more obvious.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j</artifactId>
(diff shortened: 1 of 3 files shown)


michael-simons (approved):

mp911de (approved):
Looks good.

meistermeier:
Merged with c72ecc7b5f73290062a1a8b633923f9a5ee96527 (master) and b395d22a39caab9914d97f38da31fd59d1cfe144 (6.0)

PR #565, opened 2020-12-16
Fix bidirectional relationship properties persistence.

meistermeier:
Previously a bidirectional relationships would have run into a StackOverflow on save.
We check for the existence in the already processed relationships `stateMachine` but those relationships with properties get wrapper objects created each time on access.
This led to the situation that, even though the very same relationship were processed, the exit condition were never met.

Short version: I have overwritten `hashCode` and `equals` 😺  Not even by myself but told the IDE to do so.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j</artifactId>
-	<version>6.1.0-SNAPSHOT</version>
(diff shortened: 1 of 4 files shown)


michael-simons (approved):
Please merge in 6.1 and 6.0. :)

meistermeier:
Fixed with 1ea1c325374b7e2df07633b6ed72b87605e3ce9e (master) and 810af7b9fe681aec5a94c7010fb4bfbb61b8d4ff (6.0.x)

PR #2101, opened 2021-01-07
Correctly fetch properties and relationships in hierarchy

### LINKED ISSUE #2028: Relationship is always null with inheritance [DATAGRAPH-1467]
**[Julian](https://jira.spring.io/secure/ViewProfile.jspa?name=JIRAUSER50788)** opened **[DATAGRAPH-1467](https://jira.spring.io/browse/DATAGRAPH-1467?redirect=false)** and commented

I'm having trouble mapping a relationship which is involved into an Entity with inheritance.

I'm trying to figure this out for three days now but no luck so far. Regardless of my parameter combination, if I create a no args constructor, an all args constructor the relationship is not mapped. For version 6.0.2 the relationship was always populated as null. Interestingly when I define a Collection of Posts in Tag, since version 6.1 M1 this works. Versions prior to this did not work. Now comes the weird thing: versions prior to 6.1 M1 work, when I define a Post Collection with Relationship annotation in Tag, but the Relationship Type has to be some sort of different non existing relationship. Regardless of that the query gets really slow (one second plus). Also for obvious reasons I don't want to populate my Tag with a Post Collection.

I tried moving the Relationship Collection to each individual sub entity, which also did not yield the desired behavior. At some point I thought okay, maybe its just Kotlin being Kotlin, and recreated the situation in Java as well -> same behavior. I attached a demo project which illustrates my problem even further. In the mean time here is also some Kotlin code which some what describes the problem. In the demo.zip file is also a sample data CSV export from neo4j.

Thanks in advance, contact me if you need further elaboration.

 

```java
@Node("Tag")
data class Tag(@Id @Property(name = "id") val id: String)

@Node("Post")
abstract class Post(
    @Id @GeneratedValue val id: UUID? = null,
    @Relationship(type = "BELONGS_TO", direction = OUTGOING) val tags: Set<Tag>
)

@Node("TextPost")
class TextPost(tags: Set<Tag>, val text : String) : Post(tags = tags)

@Node("ImagePost")
class ImagePost(tags: Set<Tag>, val image : String) : Post(tags = tags)
```

---

**Affects:** 6.1 M1 (2021.0.0), 6.0.2 (2020.0.2)

**Attachments:**
- [Application.zip](https://jira.spring.io/secure/attachment/28862/Application.zip) (_778.34 kB_)
- [demo.zip](https://jira.spring.io/secure/attachment/28861/demo.zip) (_778.34 kB_)
- [demo.zip](https://jira.spring.io/secure/attachment/28859/demo.zip) (_768.57 kB_)

### LINKED ISSUE #2031: InstantiationError for Abstract Nodes [DATAGRAPH-1470]
**[Joshua Mims](https://jira.spring.io/secure/ViewProfile.jspa?name=JIRAUSER50805)** opened **[DATAGRAPH-1470](https://jira.spring.io/browse/DATAGRAPH-1470?redirect=false)** and commented

Here is the basic setup:

 

```java
@Data
@Node
public class Person {  
    
    @Property 
    private String name;
    
    @Relationship(type = "HAS_PET")
    private List<Pet> pets;
}

@Data
@Node
public abstract class Pet {
    
    @Property 
    private String name;
}

@Data
@Node(labels = { "Dog", "Pet" })
public class Dog extends Pet {  

    @Property
    private String breed;
}

@Repository
public interface PetRepository extends Neo4jRepository<Pet, Long> {  

    @Query("MATCH (pet:Pet) WHERE pet.name = $name RETURN pet")
    findByName(String name);

    @Query("MATCH (pet:Pet)<-[has_pet:HAS_PET]-(person:Person) WHERE pet.name = $name RETURN pet, has_pet, person")
    findByNameWithOwner(String name);
}
```

Here we have some polymorphism.  `PetRepository::findByName` will return correctly, and the `Pet` node return should be an instance of `Dog`.  However, `PetRepository::findByNameWithOwner` will throw an InstantiationError for `Pet` (because `Pet` is abstract and cannot be instantiated).  After doing much digging into the framework, I believe this is because, in this instance, we are returning more than one object aside from the root ('Pet'), and in this case, `DefaultNeo4jEntityConverter::mergeRootNodeWithRecord` is called.  After this, the previous `MapAccessor` information is lost (most importantly, the `labels` property).  When this is passed on as `queryRoot` to the `map` method to later instantiate the object, it fails since there is no concrete type information left to create an instance from, and what's left is the abstract type

---

**Affects:** 6.0.2 (2020.0.2)

meistermeier:
This closes #2028 and _could_ also close #2031


michael-simons (approved):
I am smiling.

Issue #2114, opened 2021-01-14
Include all nodes of a path during mapping.


Jafeyyu:
https://github.com/spring-projects/spring-data-neo4j/blob/f18e15a6b899cdcdf719e1bc68e61c4c4a7d7d3d/src/main/java/org/springframework/data/neo4j/core/mapping/DefaultNeo4jEntityConverter.java#L487
The Nodes of Path need to filter should be PATH_START and PATH_END,but there is PATH_START and PATH_START

michael-simons:
🤦 Oh dear.

Thank you so much.

michael-simons:
There are a couple of scenarios where not all nodes or not all relationships of a path have been included in the mapping. First of call we filtered on the wrong path segment as documented by @Jafeyyu in this ticket and in some scenarios the synthesization of values did not include the segments in such a way that everything in the result was actually used in the mapping.

This is fixed by synthesizing a result that has the form as a custom query (root node, all relationships and all related nodes), instead of going through the paths.

Jafeyyu:
I think you are right, I thought so before, but I don't know how to describe it, because my English is not very good, because of this problem, I put the version at 5.6.3 in my project,Thank you for fixed it

michael-simons:
No worries, your contribution is appreciated, I wouldn't have the issue so soon without it. Take care!

PR #2115, opened 2021-01-14
Change the nodes to be filtered by targetLabel

Jafeyyu:
Change the nodes to be filtered by targetLabel from PATH_START&PATH_START to PATH_START&PATH_END.

- [x] You have read the [Spring Data Neo4j contribution guidelines](https://github.com/spring-projects/spring-data-neo4j/blob/master/CONTRIBUTING.adoc).
- [ ] You submit test cases (unit or integration tests) that back your changes.
- [x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).


michael-simons:
Thanks a lot, @Jafeyyu As much as I would have liked to merge or cherry pick that into my branch with tests added, I cannot.

Of course I had that solution myself, but it turned out that we can simplify the path mapping by a magnitude without changing functionality. You will see in the solution to #2114.

PR #2122, opened 2021-01-19
Optimize mapping and querying for cyclic data

### LINKED ISSUE #2121: Optimize path querying and mapping
The path-based querying and mapping should result in records that can be processed as:
variable | description
-------- | -----------
n | root / leading node of the record
__srn__ | `collect(nodes(allPathSegments))`
__sr__ | `collect(relationships(allPathSegments))`

meistermeier:
Crunch down the path(s) / path segments in the result to distinct lists of relationships and nodes on the database side to process them in a unified way.

Closes #2121


meistermeier:
Closed with commits f9e7896 (master) and cc73484 (6.0)

PR #2161, opened 2021-02-19
Push database selection down to (Reactive)Neo4jClient

meistermeier:
This was done to give also users of the pure (Reactive)Neo4jClient the option to use the `DatabaseSelectionProvider` (or its reactive pendant).

We might have to discuss if the bean creation methods in the `AbstractNeo4jConfig` and `AbstractReactiveNeo4jConfig` need to keep the signature with the unused `DatabaseSelectionProvider` / `ReactiveDatabaseSelectionProvider` for a while with deprecation warnings and a replacement following in up in 6.2.0.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j</artifactId>
-	<version>6.1.0-SNAPSHOT</version>
+	<version>6.1.0-GH-2159-SNAPSHOT</version>
 
 	<name>Spring Data Neo4j</name>
 	<description>Next generation Object-Graph-Mapping for Spring Data.</description>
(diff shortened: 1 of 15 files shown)


michael-simons (changes requested):
What we forgot here is that this needs to work together with a Spring Boot change: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jDataAutoConfiguration.java#L87-L114 That change alone here won't make the users happy.

meistermeier:
But this can only be done if the changes are merge, or?

meistermeier:
Merged with 210f209e1afd82300b74870aed7df9360a38309b

PR #2167, opened 2021-03-04
Respect class hierachy when querying for and mapping to generic classes

meistermeier:
This solves two problems that go hand in hand with each other:
a) The created query did not respect potential relationships in the child classes.
b) The mapping (`Neo4j..Converter`) maps a graph 
`(A)-(:B:A)-(:C:A)`
as 
`(A)-(:B:A)-(:B:A)`
because it keeps the same concrete node description of the prior mapped object in the chain if the next one has the same base class (`A`).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j</artifactId>
-	<version>6.0.6-SNAPSHOT</version>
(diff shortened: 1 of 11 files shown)


michael-simons (approved):

meistermeier:
Merged with dbf28598955b021c103cc94c5f0de6017cd29a69 (master) and e7929c21b4d063f321405469c316ccb6f3de5113 (6.0.x)

PR #2173, opened 2021-03-11
Create new objects with id set when saving.

meistermeier:


src/main/java/org/springframework/data/neo4j/core/ReactiveNeo4jTemplate.java
```diff
@@ -562,140 +566,224 @@ private Object convertIdValues(@Nullable Neo4jPersistentProperty idProperty, Obj
...
+			Flux<RelationshipPropertyValues> relationshipCreation = Flux.fromIterable(relatedValuesToStore).flatMap(relatedValueToStore -> {
+
+				Object relatedNodePreEvt = relationshipContext.identifyAndExtractRelationshipTargetNode(relatedValueToStore);
+				return Mono.deferContextual(ctx -> eventSupport.maybeCallBeforeBind(relatedNodePreEvt)
+						.flatMap(relatedNode -> Mono.just(neo4jMappingContext.getPersistentEntity(relatedNodePreEvt.getClass()))
+								.flatMap(targetEntity -> Mono.just(targetEntity.isNew(relatedNode))
+										.flatMap(isNew -> saveRelatedNode(relatedNode, relationshipContext.getAssociationTargetType(),
+												targetEntity)
+												.flatMap(relatedInternalId -> {
+													// if an internal id is used this must get set to link this entity in the next iteration
```
michael-simons, line 657: > get set

?

src/main/java/org/springframework/data/neo4j/core/ReactiveNeo4jTemplate.java
```diff
@@ -562,140 +566,224 @@ private Object convertIdValues(@Nullable Neo4jPersistentProperty idProperty, Obj
...
+																if (processState != ProcessState.PROCESSED_ALL_VALUES) {
+																	something = processNestedRelations(targetEntity, targetPropertyAccessor,
+																			isNew, stateMachine);
+																} else {
+																	something = Mono.just((T) targetPropertyAccessor.getBean());
+																}
+
+																return something.flatMap(currentValue -> {
+																	Object newRelationshipObject = targetPropertyAccessor.getBean();
+																	if (relationshipDescription.hasRelationshipProperties()) {
```
michael-simons, line 694: Interesting choice for a variable name.

src/main/java/org/springframework/data/neo4j/core/ReactiveNeo4jTemplate.java
```diff
@@ -562,140 +566,224 @@ private Object convertIdValues(@Nullable Neo4jPersistentProperty idProperty, Obj
...
+																		Object lalelu = relationshipProperty.isDynamicAssociation()
```
michael-simons, line 695: I would rather prefer to bite it and deduplicate this conditions, similar on what we did together with the checks for cyclic dependencies, regardless of the fact that one lives in the imperative world and the other in the reactive world.

src/test/java/org/springframework/data/neo4j/integration/imperative/RepositoryIT.java
```diff
@@ -1802,20 +1801,6 @@ void createWithCustomQueryShouldWorkWithPlainObjects(@Autowired PersonRepository
 			assertThat(newPerson.getBornOn()).isEqualTo(p.getBornOn());
 		}
 
-		@Test // GH-2141
-		void saveWithGeneratedIdsReturnsObjectWithIdSet(
```
michael-simons, line 1806: I appreciate that removal a lot.

michael-simons (changes requested):
In general I am wondering if there's a way to check upfront wether the recreation of the collections is actually necessary (at least if I don't understand the changes wrong, the templates seem to create at least the new approximate collection in all cases, regardless of whether working with immutable or not.).

meistermeier:
Merged with f9267d35a

PR #2179, opened 2021-03-13
Fix Pageable and Sort for cyclic domains

meistermeier:
`Sort` and (in non-reactive world) `Pageable` in combination with cyclic domains and a `findAll` call led to a runtime exception because the queries were only partial created.


src/main/java/org/springframework/data/neo4j/repository/query/QueryFragmentsAndParameters.java
```diff
@@ -260,6 +275,33 @@ public boolean isScalarValueReturn() {
 			return orderBy != null ? orderBy : new SortItem[]{};
 		}
 
+		public Statement toGenericStatement() {
```
michael-simons, line 278: Maybe something like "generateGenericStatement" conveys what's happening better. And maybe a comment for me.

Also: Does it need to be public?

michael-simons (changes requested):
Basically just naming changes…

meistermeier:
Merged with 1a6a2f336 and d84991b3b

PR #2181, opened 2021-03-14
Make state machine for saving process more robust.

meistermeier:
If two entities and one were saved as
`(:A{name:'a1'})-[:KNOWS]->(:A{name:'a2'})`
and we would add another back-referencing `KNOWS` relationship from `a2` to `a1` in a clean operation
`loadA1` (which automatically pulls also `a2`)
`loadA2`
`createA2KnowsA1`
the state machine assumes that a`a2` from the existing `a1` relationship differs from the directly loaded `a2`
and wants to re-create it which causes in the scenario of activated optimistic locking an `OptimisticLockingException` because "one of the two" `a2`s already got an update of its version.
To mitigate the problem, a registration of initial objects in the state machine to mark them as processed were introduced.

Besides that the state machine got a little bit more robust, I wonder if the approach to check for objects' equality, that results in a `PROCESSED_ALL_VALUES` state, is the right one. I had to adjust one existing test because there were two "different" objects that were "equal" due to a very simple equals method.

### DIFF
--- a/src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
+++ b/src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
@@ -565,9 +565,6 @@ public final class Neo4jTemplate implements Neo4jOperations, BeanFactoryAware {
 		Neo4jPersistentProperty requiredIdProperty = targetNodeDescription.getRequiredIdProperty();
 		PersistentPropertyAccessor<Object> ding = targetNodeDescription.getPropertyAccessor(entity);
 		Object idValue = ding.getProperty(requiredIdProperty);
-		if (idValue == null) {
-			System.out.println("nein");
-		}
 
 		return neo4jClient.query(() ->
 				renderer.render(cypherGenerator.prepareMatchOf(targetNodeDescription,
--- a/src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
+++ b/src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
@@ -323,8 +323,11 @@ public final class Neo4jTemplate implements Neo4jOperations, BeanFactoryAware {
 				.bind(entityList).to(Constants.NAME_OF_ENTITY_LIST_PARAM).run();
 
 		// Save related
-		entitiesToBeSaved.forEach(entityToBeSaved -> processRelations(entityMetaData, entityToBeSaved,
-				isNewIndicator.get(entitiesToBeSaved.indexOf(entityToBeSaved)), databaseName, entities.get(entitiesToBeSaved.indexOf(entityToBeSaved))));
+		entitiesToBeSaved.forEach(entityToBeSaved -> {
+			int positionInList = entitiesToBeSaved.indexOf(entityToBeSaved);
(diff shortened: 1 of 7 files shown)


src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
```diff
@@ -553,6 +559,23 @@ public void deleteAll(Class<?> domainType) {
 		return (T) propertyAccessor.getBean();
 	}
 
+	private <Y> Long queryRelatedNode(Object entity, Neo4jPersistentEntity<?> targetNodeDescription,
+									  @Nullable String inDatabase) {
+
+		Neo4jPersistentProperty requiredIdProperty = targetNodeDescription.getRequiredIdProperty();
+		PersistentPropertyAccessor<Object> ding = targetNodeDescription.getPropertyAccessor(entity);
```
michael-simons, line 566: Naming.

src/main/java/org/springframework/data/neo4j/core/ReactiveNeo4jTemplate.java
```diff
@@ -701,6 +709,23 @@ private Object convertIdValues(@Nullable Neo4jPersistentProperty idProperty, Obj
 		});
 	}
 
+	private <Y> Mono<Long> queryRelatedNode(Object entity, Neo4jPersistentEntity<?> targetNodeDescription,
+									  @Nullable String inDatabase) {
+
+		Neo4jPersistentProperty requiredIdProperty = targetNodeDescription.getRequiredIdProperty();
+		PersistentPropertyAccessor<Object> ding = targetNodeDescription.getPropertyAccessor(entity);
```
michael-simons, line 716: Naming.

src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
```diff
@@ -284,9 +284,9 @@ private Object convertIdValues(@Nullable Neo4jPersistentProperty idProperty, Obj
 
 		String databaseName = getDatabaseName();
 
-		Collection<T> entities;
+		List<T> entities;
```
michael-simons, line 287: not that I mind, but is it necessary?

src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
```diff
@@ -324,7 +324,7 @@ private Object convertIdValues(@Nullable Neo4jPersistentProperty idProperty, Obj
 
 		// Save related
 		entitiesToBeSaved.forEach(entityToBeSaved -> processRelations(entityMetaData, entityToBeSaved,
-				isNewIndicator.get(entitiesToBeSaved.indexOf(entityToBeSaved)), databaseName));
+				isNewIndicator.get(entitiesToBeSaved.indexOf(entityToBeSaved)), databaseName, entities.get(entitiesToBeSaved.indexOf(entityToBeSaved))));
```
michael-simons, line 327: k, i See. But please change to only one `indexOf` operation.

michael-simons (changes requested):

meistermeier:
Closed with 6aef5d9cd and 65ee2ae3d

Issue #1994, opened 2020-11-16
Build is failing and PR says "All checks have passed." [DATAGRAPH-1432]


spring-projects-issues:
**[Michael Simons](https://jira.spring.io/secure/ViewProfile.jspa?name=michael.simons)** opened **[DATAGRAPH-1432](https://jira.spring.io/browse/DATAGRAPH-1432?redirect=false)** and commented

Please see this PR https://github.com/spring-projects/spring-data-neo4j/pull/554

 

It fails due to various causes and Jenkins actually is aware https://jenkins.spring.io/view/SpringData/job/spring-data-neo4j/job/issue%252FDATAGRAPH-1428/4/

 

The Pr check should not indicate that things are good

---
No further details from [DATAGRAPH-1432](https://jira.spring.io/browse/DATAGRAPH-1432?redirect=false)

michael-simons:
Ping.

michael-simons:
I am unsure if this is solved or not but I do want to get the ticket out of my sight… I assume the Jenkins ping to GH works.

PR #2184, opened 2021-03-15
Provide a way for defining dynamic queries in a type safe way.

michael-simons:
This PR depends on the release and update of Cypher-DSL 2021.1.0 and is also missing docs atm, but I would like to have some feedback.


src/main/java/org/springframework/data/neo4j/core/mapping/Constants.java
```diff
@@ -28,7 +28,7 @@
  * @soundtrack Milky Chance - Sadnecessary
  * @since 6.0
  */
-@API(status = Status.INTERNAL, since = "6.0")
+@API(status = Status.EXPERIMENTAL, since = "6.0")
```
meistermeier, line 31: Any reasons for this change?

src/main/java/org/springframework/data/neo4j/repository/query/QueryFragmentsAndParameters.java
```diff
@@ -161,12 +194,7 @@ static QueryFragmentsAndParameters forExample(Neo4jMappingContext mappingContext
 		queryFragments.setReturnExpressions(returnStatement);
 
 		if (pageable != null) {
-			Sort pageableSort = pageable.getSort();
-			long skip = pageable.getOffset();
-			int pageSize = pageable.getPageSize();
-			queryFragments.setSkip(skip);
-			queryFragments.setLimit(pageSize);
-			queryFragments.setOrderBy(CypherAdapterUtils.toSortItems(entityMetaData, pageableSort));
+			adaptPageable(entityMetaData, pageable, queryFragments);
```
meistermeier, line 197: Spoiler: Here will be merge fun.

meistermeier (approved):
Looks good to me. Just a question and an information.

src/main/java/org/springframework/data/neo4j/core/mapping/Constants.java
michael-simons: So that people can use that in their queries to refer to the root object.

michael-simons:
@gregturn This PS cannot be green. are there any updates on this https://github.com/spring-projects/spring-data-neo4j/issues/1994

Why can't this be green? Because it depends on a snapshot dependency that is no where available.

michael-simons:
See https://jenkins.spring.io/blue/organizations/jenkins/spring-data-neo4j/detail/issue%2FGH-1911/10/pipeline#step-25-log-247

michael-simons:
Merged with f2a0383ad754965cf8c2fca1206896eaac49e49b.

PR #2195, opened 2021-03-22
Fix state machine for more complex scenarios.

meistermeier:
Two things are solved with this:
1. Check if the related value was already processed instead of checking for all values
2. Only trigger the delete of relationships if the relationship description was not processed before.


michael-simons (approved):

meistermeier:
Merged with 319a84de9 (master) and 1823b9d3d (6.0.x)

PR #2198, opened 2021-03-23
Don't override relationships if they are different.

meistermeier:


src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
```diff
@@ -531,18 +531,27 @@ public void deleteAll(Class<?> domainType) {
...
+							.to(Constants.FROM_ID_PARAMETER_NAME) //
 						.bind(relatedInternalId) //
 							.to(Constants.TO_ID_PARAMETER_NAME) //
+						.bind(idValue) //
+							.to(Constants.NAME_OF_KNOWN_RELATIONSHIP_PARAM) //
 						.bindAll(statementHolder.getProperties())
 						.fetchAs(Long.class).one();
 
-				if (idProperty != null) {
+				if (idProperty != null && isNewRelationship) {
```
michael-simons, line 554: Makes sense in the Cypher `prepareSaveOfRelationshipWithProperties`, here I am wondering a bit if that should be `|| isNewRelationship`

meistermeier:
Merged with 9944cf8e4 (master) and 91dd0851a (6.0.x)

PR #2242, opened 2021-04-27
Stabilize immutable mapping

meistermeier:
* Register aliases for mapped objects in a cache (before -> after) to avoid duplicates.
* Cache internal identifier for the relationship creation phase.
* Something with otter

### DIFF
--- a/src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
+++ b/src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
@@ -245,15 +245,21 @@ public final class Neo4jTemplate implements Neo4jOperations, BeanFactoryAware {
 				.with(neo4jMappingContext.getRequiredBinderFunctionFor((Class<T>) entityToBeSaved.getClass()))
 				.fetchAs(Long.class).one();
 
-		if (entityMetaData.hasVersionProperty() && !optionalInternalId.isPresent()) {
-			throw new OptimisticLockingFailureException(OPTIMISTIC_LOCKING_ERROR_MESSAGE);
+
+		if (!optionalInternalId.isPresent()) {
+			if (entityMetaData.hasVersionProperty()) {
+				throw new OptimisticLockingFailureException(OPTIMISTIC_LOCKING_ERROR_MESSAGE);
+			}
+			// defensive exception throwing
+			throw new IllegalStateException("Could not retrieve an internal id while saving.");
 		}
 
+		Long internalId = optionalInternalId.get();
 		PersistentPropertyAccessor<T> propertyAccessor = entityMetaData.getPropertyAccessor(entityToBeSaved);
 		if (entityMetaData.isUsingInternalIds()) {
-			propertyAccessor.setProperty(entityMetaData.getRequiredIdProperty(), optionalInternalId.get());
+			propertyAccessor.setProperty(entityMetaData.getRequiredIdProperty(), internalId);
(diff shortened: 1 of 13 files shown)


src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
```diff
@@ -244,15 +245,21 @@ private Object convertIdValues(@Nullable Neo4jPersistentProperty idProperty, Obj
...
 				.fetchAs(Long.class).one();
 
-		if (entityMetaData.hasVersionProperty() && !optionalInternalId.isPresent()) {
-			throw new OptimisticLockingFailureException(OPTIMISTIC_LOCKING_ERROR_MESSAGE);
+
+		if (!optionalInternalId.isPresent()) {
+			if (entityMetaData.hasVersionProperty()) {
+				throw new OptimisticLockingFailureException(OPTIMISTIC_LOCKING_ERROR_MESSAGE);
+			}
+			// defensive exception throwing
```
michael-simons, line 253: ?

src/main/java/org/springframework/data/neo4j/core/ReactiveNeo4jTemplate.java
```diff
@@ -256,7 +256,7 @@ private Object convertIdValues(@Nullable Neo4jPersistentProperty idProperty, Obj
 						if (entityMetaData.isUsingInternalIds()) {
 							propertyAccessor.setProperty(entityMetaData.getRequiredIdProperty(), internalId);
 						}
-					}).then(Mono.defer(() -> processRelations(entityMetaData, instance, propertyAccessor, inDatabase, isNewEntity)));
+					}).flatMap(internalId -> processRelations(entityMetaData, instance, internalId, propertyAccessor, inDatabase, isNewEntity));
```
michael-simons, line 259: That changes semantics… It will process relations now in parallel to the original publisher.

src/test/java/org/springframework/data/neo4j/integration/reactive/ReactiveImmutableExternallyGeneratedIdsIT.java
```diff
@@ -70,8 +85,7 @@ void saveWithExternallyGeneratedIdsReturnsObjectWithIdSet(
 				.verifyComplete();
 	}
 
-	@Test
-		// GH-2141
+	@Test // GH-41
```
michael-simons, line 88:
```suggestion
	@Test // GH-2141
```

michael-simons (changes requested):

src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
meistermeier: The optional value of the internal id should always exist besides that the optimistic locking fails.

src/main/java/org/springframework/data/neo4j/core/ReactiveNeo4jTemplate.java
meistermeier: Should this then be wrapped into a `flapMap(internalId -> Mono.defer()....`?
I thought the same but found the other option not really appealing and things are still working as expected.

meistermeier:
Merged with 
fd36b91b8 (main)
6fde45cd5 (6.1.x)
d846c3543 (6.0.x)

PR #2253, opened 2021-05-11
Add support for distinct derived queries

meistermeier:
The addition of `distinct` is only needed for queries that do not work with a circular model / projection.
I tried to come up with other examples beside the obvious one. Maybe you can chime in to challenge the very simplistic logic ;)


michael-simons (approved):
Nice.

meistermeier:
Closed with 08caa54f1 (main), e812631c7 (6.1.x) and 7cca8c189 (6.0.x)

Issue #1907, opened 2020-07-27
Would be nice to have @OrderBy on collection relationships [DATAGRAPH-1345]


spring-projects-issues:
**[Gerrit Meier](https://jira.spring.io/secure/ViewProfile.jspa?name=gerrit.meier)** opened **[DATAGRAPH-1345](https://jira.spring.io/browse/DATAGRAPH-1345?redirect=false)** and commented

It would be nice to have `@OrderBy` annotation specifying the ordering of the elements of a `@Relationship` collection.

For example:

 

{{`@OrderBy`("firstName ASC, lastName ASC")
`@Relationship`("HAVE_USER")
private Set\<User> users = new LinkedHashSet<>();}}

This would perform the ordering at the DB level by firstName then lastName with ASC. I know that we can do it programmatically with SortedSet but I guess DB level is a much more great way of doing this.

A `LinkedHashSet` preserves order based on insertion order that is why I used that one

---

**Affects:** 6.0 M2 (2020.0.0)

**Reference URL:** https://github.com/neo4j/sdn-rx/issues/227

michael-simons:
Hi @aaramg while this is a quite relatable and good idea, we cannot implement it as such.
We do use a thing called [List comprehensions](https://neo4j.com/docs/cypher-manual/current/syntax/lists/) to fetch the relationships. Those cannot be sorted inline and we would have to unroll them first after creating them, defeating their purpose.

My suggestion is implementing orderable on your domain object and using a sorted collection. We do reuse the collection type.

PR #2258, opened 2021-05-18
Clean-up relationships querying code.

michael-simons:
Investigating #1907 provided a bit of opportunity to clarify the relationship mapping code a bit when deciding to use generic "find by id" queries vs. list projections. 

While we can't implement 1907 as such, I would like to see this refactorings merged in main and 6.1


src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
```diff
@@ -410,9 +412,9 @@ private Object convertIdValues(@Nullable Neo4jPersistentProperty idProperty, Obj
 		}
 
 		class Tuple3<T> {
-			T t1;
-			boolean t2;
-			T t3;
+			final T t1;
+			final boolean t2;
```
meistermeier, line 416: Because it got touched again, can we please introduce some proper names for the fields in the tuple?

meistermeier (approved):

PR #2291, opened 2021-06-15
Simplify relationship determination logic.

meistermeier:
Moved the `....findFirst` logic into the constructor mapping because here we can be sure that there is only one field with the same name in the same class.


michael-simons (approved):
I wonder if this is still valid after the other fix, but looks good to me. 
If still needed, please merge and carry forward to 6.1. and main.

meistermeier:
Closed by 
cdef2fa60 (main) 
10398ea93 (6.1.x)
e1ea132c9 (6.0.x)

PR #2312, opened 2021-06-30
Allow inheritance with concrete classes in the chain.

meistermeier:


michael-simons (approved):

meistermeier:
Thanks. There is still something unstable. I will merge this after the last flaky behaviour is gone ;)

meistermeier:
Merged with ea28dcade (main) and 0c0de1b31 (6.1.x)

PR #2314, opened 2021-07-02
Multi-level projection support

meistermeier:


meistermeier:
Merged by 408f2d004 (main) and cbe85d87e (6.1.x)

PR #2319, opened 2021-07-09
Mapping performance optimization

meistermeier:


meistermeier:
Merged with 6d1b23e22 (main) and e5de74464 (6.1.x)

PR #2341, opened 2021-07-26
Use domainClass as driver for the root node name.

meistermeier:


src/main/java/org/springframework/data/neo4j/core/mapping/Constants.java
```diff
@@ -31,7 +34,10 @@
 @API(status = Status.EXPERIMENTAL, since = "6.0")
 public final class Constants {
 
-	public static final SymbolicName NAME_OF_ROOT_NODE = Cypher.name("n");
```
michael-simons, line 34: I would do a `NAME_OF_TYPED_ROOT_NODE.apply(null)` and assigned that, so that it is somewhat compatible in case anyone used those constants.

src/test/java/org/springframework/data/neo4j/test/TestConstants.java
```diff
@@ -0,0 +1,30 @@
...
+import org.neo4j.cypherdsl.core.SymbolicName;
+
+/**
+ * Constants for test purposes only.
+ *
+ * @author Gerrit Meier
+ */
+public class TestConstants {
+
+	public static final SymbolicName NAME_OF_ROOT_NODE = Cypher.name("n");
```
michael-simons, line 28: that whole class can go away if you agree with me on the first suggestion.

michael-simons (changes requested):

michael-simons (approved):
👍

meistermeier:
Merged with 44185a4150b6a2682fcde122adc3ed1ea55a875b

PR #2366, opened 2021-08-30
Fix hydration problem with sliced data

meistermeier:


michael-simons (changes requested):
Not happy with the logic in `populateFrom`. I think I get what it is going, but I would like to read more easy… like "this target collection already exists, I can add things to it, but I cannot create the target again"

michael-simons (approved):

meistermeier:
Merged with 6db017846 (main) and dd174f7b7 (6.1.x)

PR #2399, opened 2021-10-08
Fix flaky RelationshipProperties field determination.

meistermeier:


meistermeier:
Merged with c458bc05b (7.0.x) / fdc914ac7 (main) / 71a512990 (6.1.x)

PR #2422, opened 2021-11-02
Add support for user impersonation.

michael-simons:
This is ready for review / polishing and should go in as soon as possible, even without the latest driver (we use reflection to set the selected user).

### DIFF
--- a/src/main/java/org/springframework/data/neo4j/core/transaction/Neo4jTransactionContext.java
+++ b/src/main/java/org/springframework/data/neo4j/core/transaction/Neo4jTransactionContext.java
@@ -17,8 +17,10 @@ package org.springframework.data.neo4j.core.transaction;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Objects;
 
 import org.neo4j.driver.Bookmark;
+import org.springframework.data.neo4j.core.ImpersonatedUser;
 import org.springframework.lang.Nullable;
 
 /**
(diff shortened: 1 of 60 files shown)


src/main/java/org/springframework/data/neo4j/core/transaction/Neo4jTransactionContext.java
```diff
@@ -34,28 +35,48 @@
 	/**
 	 * The target database of the session.
 	 */
-	private @Nullable final String databaseName;
+	private final DatabaseSelection databaseSelection;
+
+	/**
+	 * Will be {@literal null} for the default user, non-null for an impersonated user.
```
meistermeier, line 41: Doesn't it default to `CONNECTED_USER` instead of `null`?

meistermeier (changes requested):

src/main/java/org/springframework/data/neo4j/core/transaction/ReactiveNeo4jTransactionManager.java
```diff
@@ -47,6 +50,75 @@
...
+
+		/**
+		 * Configures a provider for impersonated users. Make sure to use the same instance as for a possible
+		 * {@link org.springframework.data.neo4j.core.ReactiveNeo4jClient}. During runtime, it will be checked if a call is made
+		 * for the same user when happening in a managed transaction.
+		 *
+		 * @param userSelectionProvider The provider for impersonated users
+		 * @return The builder
+		 */
+		public Builder withUserProvider(@Nullable ReactiveUserSelectionProvider userSelectionProvider) {
```
meistermeier, line 107:
```suggestion
		public Builder withUserSelectionProvider(@Nullable ReactiveUserSelectionProvider userSelectionProvider) {
```

meistermeier (changes requested):

michael-simons:
Closed with c5e68a4

Issue #2420, opened 2021-11-02
Allow dynamic projection for Neo4jTemplate::save


Andy2003:
There are situations when you want to dynamically determine which fields of a domain object should actually be saved.

To achieve this it would be good to have a method similar to `Neo4jTemplate::saveImpl` available as a public API

https://github.com/spring-projects/spring-data-neo4j/blob/ea7bfb71b1669ca6e88fc9a06f48fb90f0c21739/src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java#L360-L411

Andy2003:
Is there a possibility of downporting this to 6.2 so we don't have to wait for the next spring boot release?

PR #2454, opened 2021-11-30
Provide `saveAs` overloads with a predicate for selecting projected attributes.

michael-simons:
This implements #2420. Please @Andy2003 check if this suites your needs. 

Before you ask: Chances are minimal to have something similar for reading. For read it just feels like the wrong approach. If we are down there, one might ask for the relevance of the mapping process itself and just delegate to the client.

Thanks @meistermeier for your help on this. I did not touch anything that might be affected by https://github.com/spring-projects/spring-data-commons/pull/2420

Funny coincidence that those things have the same issue number :D


src/main/java/org/springframework/data/neo4j/core/mapping/PropertyTraverser.java
```diff
@@ -0,0 +1,95 @@
...
+	) {
+		traverse(root, (path, toProperty) -> true, sink);
+	}
+
+	public synchronized void traverse(
+			Class<?> root,
+			BiPredicate<PropertyPath, Neo4jPersistentProperty> predicate,
+			BiConsumer<PropertyPath, Neo4jPersistentProperty> sink
+	) {
+		this.pathsTraversed.clear();
```
meistermeier, line 60: Given this call and the current usage looks like `new PropertyTraverser(..).traverse(..)` wouldn't it make sense to make those methods _static_?

src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
```diff
@@ -451,10 +462,10 @@ private Object convertIdValues(@Nullable Neo4jPersistentProperty idProperty, Obj
 
 	@Override
 	public <T> List<T> saveAll(Iterable<T> instances) {
-		return saveAllImpl(instances, Collections.emptyMap());
+		return saveAllImpl(instances, Collections.emptyMap(), null);
 	}
 
-	private <T> List<T> saveAllImpl(Iterable<T> instances, Map<PropertyPath, Boolean> includedProperties) {
+	private <T> List<T> saveAllImpl(Iterable<T> instances, @Nullable Map<PropertyPath, Boolean> includedProperties, @Nullable  BiPredicate<PropertyPath, Neo4jPersistentProperty> includeProperty) {
```
meistermeier, line 468:
```suggestion
	private <T> List<T> saveAllImpl(Iterable<T> instances, @Nullable Map<PropertyPath, Boolean> includedProperties, @Nullable BiPredicate<PropertyPath, Neo4jPersistentProperty> includeProperty) {
```

src/main/java/org/springframework/data/neo4j/core/ReactiveNeo4jTemplate.java
```diff
@@ -495,7 +512,7 @@ private Object convertIdValues(@Nullable Neo4jPersistentProperty idProperty, Obj
 		}).map(instance -> projectionFactory.createProjection(resultType, instance));
 	}
 
-	private <T> Flux<T> saveAllImpl(Iterable<T> instances, @Nullable Map<PropertyPath, Boolean> includedProperties) {
+	private <T> Flux<T> saveAllImpl(Iterable<T> instances, @Nullable Map<PropertyPath, Boolean> includedProperties, @Nullable  BiPredicate<PropertyPath, Neo4jPersistentProperty> includeProperty) {
```
meistermeier, line 515:
```suggestion
	private <T> Flux<T> saveAllImpl(Iterable<T> instances, @Nullable Map<PropertyPath, Boolean> includedProperties, @Nullable BiPredicate<PropertyPath, Neo4jPersistentProperty> includeProperty) {
```

src/main/java/org/springframework/data/neo4j/core/Neo4jOperations.java
```diff
@@ -191,6 +193,24 @@
...
+	 * Be careful when reusing the returned instance for further persistence operations, as it will most likely not be
+	 * fully hydrated and without using a static or dynamic projection, you will most likely cause data loss.
+	 *
+	 * @param instance        the entity to be saved. Must not be {@code null}.
+	 * @param includeProperty A predicate to determine the properties to save.
+	 * @param <T>             the type of the entity.
+	 * @return the saved instance.
+	 * @since 6.3
+	 */
+	default <T> T saveAs(T instance, BiPredicate<PropertyPath, Neo4jPersistentProperty> includeProperty) {
```
meistermeier, line 210: Since we accept it, the instance should be `@Nullable`. Would also make sense for the `saveAllAs` method and the reactive counter-parts. (Would need an additional null check in the `saveAllAs`/`saveAllImpl`).

src/main/java/org/springframework/data/neo4j/core/Neo4jOperations.java
```diff
@@ -213,6 +233,24 @@
 	 */
 	<T> List<T> saveAll(Iterable<T> instances);
 
+	/**
+	 * Saves several instances of an entity, using the provided predicate to shape the stored graph. One can think of the predicate
+	 * as a dynamic projection. If you want to save or update properties of associations (aka related nodes), you must include
+	 * the association property as well (meaning the predicate must return {@literal true} for that property, too).
+	 * <p>
+	 * Be careful when reusing the returned instance for further persistence operations, as it will most likely not be
```
meistermeier, line 241:
```suggestion
	 * Be careful when reusing the returned instances for further persistence operations, as they will most likely not be
```

src/main/java/org/springframework/data/neo4j/core/ReactiveNeo4jOperations.java
```diff
@@ -212,6 +232,24 @@
 	 */
 	<T> Flux<T> saveAll(Iterable<T> instances);
 
+	/**
+	 * Saves several instances of an entity, using the provided predicate to shape the stored graph. One can think of the predicate
+	 * as a dynamic projection. If you want to save or update properties of associations (aka related nodes), you must include
+	 * the association property as well (meaning the predicate must return {@literal true} for that property, too).
+	 * <p>
+	 * Be careful when reusing the returned instance for further persistence operations, as it will most likely not be
```
meistermeier, line 240:
```suggestion
	 * Be careful when reusing the returned instances for further persistence operations, as they will most likely not be
```

src/main/java/org/springframework/data/neo4j/core/mapping/PropertyTraverser.java
michael-simons: No. The thing has state.

Andy2003:
Given

```java
Node node2 = Node.builder()
		.name("test123")
		.parent(rootNode) // rootNode was queried before
		.build();

neo4jTemplate.saveAs(node2, (propertyPath, property) -> {
	if (propertyPath.stream().count() > 1){
		// do not save deeper than level 1
		return false;
	}
	if (!property.isRelationship()) {
		return true;
	}
	return property.getName().equals("parent");
});
```

3 write queries are triggered:

1.  the merge of the newly created node including its properties to save `RUN "MERGE (node:`Node` {nodeId: $__id__}) SET node += $__properties__ RETURN node" {__id__="5dca7a8c-8f5e-4089-aae5-5e794332b7c0", __properties__={name: "test123", nodeId: "5dca7a8c-8f5e-4089-aae5-5e794332b7c0", created: NULL}}`
2. a non required update of the root node `MERGE (node:`Node` {nodeId: $__id__}) SET node += $__properties__ RETURN node" {__id__="root", __properties__={}}`
3. The creation of the relation: `MATCH (startNode:`Node`) WHERE startNode.nodeId = $fromId MATCH (endNode) WHERE id(endNode) = $toId MERGE (startNode)-[relProps:`CHILD_OF`]->(endNode) RETURN id(relProps)" {toId=40, __knownRelationShipId__=NULL, fromId="5dca7a8c-8f5e-4089-aae5-5e794332b7c0"}`

The 2nd step should be filtered out

meistermeier:
Sorry for the late reply. Could you please describe why you explicitly add `parent` to the list but on the other hand say SDN should not persist it.
I agree, on a technical level saving a node without properties does not make any sense at first sight. But this needs also be done if e.g. the _root_ node is a new node.

Andy2003:
With `parent` I meant that the relation should be stored. But I don`t want to change any properties at the parents node.

meistermeier:
I would like to keep this out of this PR.
I am not yet convinced that we would ever make a difference between the relationship and the relationship with a related node in the future.

meistermeier:
Merged with f79ad5393a9f9a54ec800f00a71b3a87a8c257e4

PR #2566, opened 2022-07-13
Add Spring Framework AOT support.

meistermeier:
Those things that we do to support AOT 😄


src/main/java/org/springframework/data/neo4j/aot/AotNeo4jRepositoryPostProcessor.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * @author Gerrit Meier
+ */
+public class AotNeo4jRepositoryPostProcessor extends RepositoryRegistrationAotProcessor {
+
+ 	@Override
+ 	protected void contribute(AotRepositoryContext repositoryContext, GenerationContext generationContext) {
+ 		super.contribute(repositoryContext, generationContext);
+
+ 		repositoryContext.getResolvedTypes().stream().filter(Neo4jAotPredicates.IS_SIMPLE_TYPE.negate()).forEach(type -> {
```
christophstrobl, line 32: is the filter in data commons too restrictive? Which additional types are you expecting to cover here?

src/main/java/org/springframework/data/neo4j/aot/AotNeo4jRepositoryPostProcessor.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.springframework.data.neo4j.aot;
+
+import org.springframework.aot.generate.GenerationContext;
+import org.springframework.data.aot.AotRepositoryContext;
+import org.springframework.data.aot.RepositoryRegistrationAotProcessor;
+import org.springframework.data.aot.TypeContributor;
+
+/**
+ * @author Gerrit Meier
```
christophstrobl, line 24: `@since` tags would be nice to have.

src/main/java/org/springframework/data/neo4j/aot/DataNeo4jRuntimeHints.java
```diff
@@ -0,0 +1,54 @@
...
+	@Override
+	public void registerHints(RuntimeHints hints, @Nullable ClassLoader classLoader) {
+
+		hints.reflection().registerTypes(
+				Arrays.asList(
+						TypeReference.of(SimpleNeo4jRepository.class),
+						TypeReference.of(SimpleQueryByExampleExecutor.class),
+						TypeReference.of(BeforeBindCallback.class),
+						TypeReference.of(AfterConvertCallback.class),
+						// todo "temporary" fix, should get resolved when defined in @GeneratedValue
```
christophstrobl, line 46: please create an issue in spring-projects/spring-data-commons for this, so we can keep track of the progress there.

src/main/java/org/springframework/data/neo4j/aot/DataNeo4jRuntimeHints.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.data.neo4j.repository.query.SimpleQueryByExampleExecutor;
+import org.springframework.data.neo4j.repository.support.SimpleNeo4jRepository;
+import org.springframework.lang.Nullable;
+
+import java.util.Arrays;
+
+/**
+ * @author Gerrit Meier
+ */
+public class DataNeo4jRuntimeHints implements RuntimeHintsRegistrar {
```
christophstrobl, line 35: maybe just `Neo4jRuntimeHints`?

christophstrobl (approved):
left a few comments - other than that LGTM.

src/main/java/org/springframework/data/neo4j/aot/AotNeo4jRepositoryPostProcessor.java
meistermeier: TBH, I took this from the mongo commit ;)
If this is not needed, I can remove it.

christophstrobl: In that case you man want to remove the entire processor as well as the reference to it in `Neo4jRepositoryConfigurationExtension.java` and just rely on the inspection done in `data-commons`.
less code :)

PR #2697, opened 2023-03-21
Support relationships in find by example

meistermeier:
Also use the `PropertyPathWrapper`, that was only used for `PartTree` / derived queries, now for the relationship pattern creation of the find-by-example mechanism.

### DIFF
--- a/src/main/java/org/springframework/data/neo4j/repository/query/Predicate.java
+++ b/src/main/java/org/springframework/data/neo4j/repository/query/Predicate.java
@@ -22,8 +22,11 @@ import static org.neo4j.cypherdsl.core.Cypher.property;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiFunction;
 
 import org.neo4j.cypherdsl.core.Condition;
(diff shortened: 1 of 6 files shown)


src/main/java/org/springframework/data/neo4j/repository/query/Predicate.java
```diff
@@ -56,80 +62,142 @@ final class Predicate {
...
+
+			if (relationshipObject == null) {
 				continue;
 			}
 
-			boolean internalId = graphProperty.isIdProperty() && probeNodeDescription.isUsingInternalIds();
-			String propertyName = graphProperty.getPropertyName();
+			// Right now we are only accepting the first element of a collection as a filter entry.
+			// Maybe combining multiple entities with AND might make sense.
+			if (relationshipObject instanceof Collection) {
```
michael-simons, line 101:
```suggestion
			if (relationshipObject instanceof Collection collection) {
```

src/main/java/org/springframework/data/neo4j/repository/query/Predicate.java
```diff
@@ -56,80 +62,142 @@ final class Predicate {
...
+				int collectionSize = ((Collection<?>) relationshipObject).size();
```
michael-simons, line 102:
```suggestion
				int collectionSize = collection.size();
```

michael-simons (changes requested):
Just a small nitpicking, very valuable change, nice test.

PR #2733, opened 2023-06-06
Add scroll support for (Reactive)FluentQuery... classes

meistermeier:
This will add the missing support for scroll (and limit).
The API is mostly used in Spring for GraphQL but can also be invoked manually when using predicate or example executors.

Note: There is a dangling commit to add some additional filtering to the test logs. (deprecated id, unknown labels)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j</artifactId>
-	<version>7.1.1-SNAPSHOT</version>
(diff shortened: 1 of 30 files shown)


michael-simons (changes requested):
I like. A lot of nullable arguments, but I don't think it's worth changing the internal api here into a bunch of different named methods.

Waiting for the Keyset based tests ;)

michael-simons (approved):

meistermeier:
Merge with 77d94bc0a (7.1.x) and 2186367cb (main)

PR #2760, opened 2023-07-03
Remove locks in reactive bookmark related operations

meistermeier:
Cannot be merged before the driver has a fix in place. Otherwise the operations will not break in SDN but in the driver.


src/main/java/org/springframework/data/neo4j/config/AbstractNeo4jConfig.java
```diff
@@ -47,6 +48,9 @@ public abstract class AbstractNeo4jConfig extends Neo4jConfigurationSupport {
 	@Autowired
 	private ObjectProvider<UserSelectionProvider> userSelectionProviders;
 
+	@Autowired
```
meistermeier, line 51: Used field injection here (and reactive client) to avoid forced change of the `Neo4jClient` bean parameters.

michael-simons: 🙈

michael-simons (approved):

Issue #2784, opened 2023-08-28
Use id() instead of toString(id(element))


meistermeier:
In the generated Cypher statements we wanted to get a unified type back: String.
To achieve this, we made use of `elementId` for newer Neo4j versions and `toString(id(element))` for older versions.
Unfortunately the `toString` wrapping created a _very_ bad impact on the performance because an id lookup exploded into a full db scan.
To avoid this, we should still use Strings in SDN wherever possible but also do the casting on the client-side to still have the benefit of id lookups in the queries.

PR #2785, opened 2023-08-29
CypherGenerator: Use id() instead of toString(id(element))

meistermeier:
Related issue #2784 

Instead of having a unified return type (String) in the Cypher result,
SDN accepts now `Long` for id() and `String`.
The conversion will take place in the library code.

To switch between id() and elementId(), SDN will inspect the selected
CypherDSL dialect and select the needed identifier function accordingly.

### DIFF
--- a/src/main/java/org/springframework/data/neo4j/core/TemplateSupport.java
+++ b/src/main/java/org/springframework/data/neo4j/core/TemplateSupport.java
@@ -35,11 +35,15 @@ import java.util.stream.StreamSupport;
 
 import org.apiguardian.api.API;
 import org.neo4j.cypherdsl.core.Cypher;
+import org.neo4j.cypherdsl.core.FunctionInvocation;
 import org.neo4j.cypherdsl.core.Functions;
+import org.neo4j.cypherdsl.core.Named;
 import org.neo4j.cypherdsl.core.Node;
 import org.neo4j.cypherdsl.core.Relationship;
 import org.neo4j.cypherdsl.core.Statement;
+import org.neo4j.cypherdsl.core.renderer.Dialect;
 import org.neo4j.cypherdsl.core.renderer.Renderer;
+import org.neo4j.driver.Value;
 import org.neo4j.driver.types.Entity;
 import org.neo4j.driver.types.MapAccessor;
 import org.neo4j.driver.types.TypeSystem;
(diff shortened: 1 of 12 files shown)


src/main/java/org/springframework/data/neo4j/core/TemplateSupport.java
```diff
@@ -414,10 +427,35 @@ static <T> Object retrieveOrSetRelatedId(
...
+			if (driverValue.hasType(TypeSystem.getDefault().NUMBER())) {
+				return driverValue.asNumber().toString();
+			}
+			return driverValue.asString();
+		}
+
+		return value.toString();
+	}
+
+	static Object convertToLongIdOrStringElementId(Collection<String> ids) {
```
michael-simons, line 449: Why returning an Object?

src/main/java/org/springframework/data/neo4j/core/mapping/CypherGenerator.java
```diff
@@ -82,6 +84,33 @@ public enum CypherGenerator {
...
+		}
+	};
+
+	/**
+	 * Set elementIdFunction to be used for identifier function.
+	 *
+	 * @param newElementIdOrIdFunction elementId CypherDSL function
+	 * @return the modified CypherGenerator
+	 */
+	public CypherGenerator withElementIdFunction(Function<Named, FunctionInvocation> newElementIdOrIdFunction) {
```
michael-simons, line 105: Why the modified one and not a new one? If you modify it (which is ok for me), use a setter.
Also, you can leave the instance in the mapping context final if you modify it anyway.

src/main/java/org/springframework/data/neo4j/core/mapping/SpringDataCypherDsl.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.function.Function;
+
+/**
+ * Supporting class for CypherDSL related customizations.
+ *
+ * @author Gerrit Meier
+ */
+@API(status = API.Status.INTERNAL)
+public class SpringDataCypherDsl {
```
michael-simons, line 34: Make it final and add an private default constructor, just for completeness please.

michael-simons (changes requested):
Great work. Complaining on a very high level here.

src/main/java/org/springframework/data/neo4j/core/TemplateSupport.java
meistermeier: The method takes the elementId _or_ id in `String` format, because it got unified to `String` when we got it from the response. To provide it again in a subsequent query, it has to have the right type again (`Long` or `String`). Basically the callee-side calls this method and expects the "right" object type.

michael-simons: Aah yes, it will be the original list in case one fails, either heterogeneous or homogeneous. All good then.

michael-simons (approved):

meistermeier:
Merged with d208634ec (main) and 514852147 (7.1.x)

michael-simons:
Yeah!

PR #2789, opened 2023-09-08
Migrate documentation to Antora

meistermeier:
The migration to Antora does not include detailed restructuring besides putting everything on the first level that was nested within _Reference documentation_ before.

Links within the same documentation page were kept as `<<...>>` for now.


meistermeier:
merged with 670dec5

PR #2802, opened 2023-09-26
Introduce additional log for cypher notifications

meistermeier:
The notifications themself are still logged on the level, they were sent from the server.
Everything with a notification related category that currently creates a lot of noise will get logged with the `org.springframework.data.neo4j.cypher.notifications` logger.
Setting this to `error` (like https://github.com/spring-projects/spring-data-neo4j/compare/7.1.x...issue/gh-2800?expand=1#diff-4ca184e5d3f39db0cbc21d76d438877f2f39251d0768e4e9f42728c610fc408eR33) will still log general cypher related messages but not those "missing ..." warnings anymore.


michael-simons (approved):
I really appreciate this, great change and I'm totally happy to let the simple switch go :)

michael-simons (changes requested):
Can't we just use `notification.category()` to map the notifications to our logger?
If not, please move the sets away from the client into the `ResultSummaries` class.

PR #2803, opened 2023-09-27
Add missing types conversions to the documentation.

Dimibe:
Hi, 
while reading the documentation, I noticed that the section for conversions is missing some types which are supported by SDN out of the box. 

This pull request completes the documentation of the conversions using the information I took from the two classes `CypherTypes` and `AdditionalTypes`. Additionally I corrected the nonworking link to the drivers manual.

One additional thing I noticed: 
I noticed that some conversions (e.g. for int/Interger) are in the `AdditionalTypes` class, but for my understanding they would probably fit better in the `CypherTypes` class.


michael-simons (approved):
Excellent addition, also great that you fixed the link to the driver manual.

Thanks a lot and sorry for picking this up quite late.

PR #2817, opened 2023-11-02
Find the right id/element id function for assigned ids

meistermeier:
* Avoid using the outdated id function wherever possible.
* Never fall back to `toString(id((element))` (in Neo4j 4.4 scenarios)

Well, I personally hate the passing of the `canUseElementId` all over the place and am open for better suggestions.


michael-simons (approved):

meistermeier:
merged with 9ae846660 (7.1.x) and f5674afda (main)

PR #2822, opened 2023-11-17
Support ResultSummaries with Notifications with no InputPosition reported by the driver.

ellebrecht:
The Neo4j Java driver may report null for `Notification.position()`. SDN's `ResultSummaries `log formatting did not support this case yet.


michael-simons (approved):
Neat, thank you!

PR #2833, opened 2023-11-25
Extracted StatmentBuilder

akshatbiztech:
I ran DesigniteJava a smell detection tool and detected various smells. I refactored Neo4jMappingContext based on the "God Class" smell. The createstaement methods can be moved to a new class to reduce the focus of the context class.

Would love to get your feedback on this.


michael-simons:
A public, open and accessible class is neither wanted or needed here. Please see my other remarks at #2834.

PR #2829, opened 2023-11-24
Refactored processNestedRelations method.

akshatbiztech:
I ran DesigniteJava a smell detection tool and detected various smells. I refactored processNestedRelations based on the "Long Method" smell. I have tried to reduce the method length by extracting some code blocks.

Would love to get your feedback on this. 

I have attached the file that shows all the implementation smells.
[ImplementationSmells.csv](https://github.com/spring-projects/spring-data-neo4j/files/13462061/ImplementationSmells.csv)

PR #2834, opened 2023-11-25
Refactored applyFinalResultToOwner

akshatbiztech:
I ran DesigniteJava a smell detection tool and detected various smells. I refactored applyFinalResultToOwner based on the "Switch Case" smell. The switch case can be transformed into a class and used by polymorphism.

Would love to get your feedback on this.


michael-simons:
Hej @akshatbiztech thanks for reading and working with the SDN source code. I will add a couple of general remarks  first than address the PR itself 

A couple of things: We do use Sonar with SDN (and other projects) and we do accept some smells here, especially complexity and method size. Well in general it is of course correct to aim for concise methods as possible, but in a complex, general purpose object mapper this is not always possible and will eventually lead to more and not less strain.

Next: While here this is not the case, huge refactorings like #2829 impose a lot of additional mental load: We, not you, are obliged to maintain this. We do know the structure and we do know why we implemented it this way. Changing it for no good reasons apart from "more clean" leads to less orientation. And, actually worse, we would need to backport this to all maintained branches, because otherwise back porting actual bug fixes and new features will get harder.

Last: To avoid disappointment, please ask before undergoing such an endeavour by opening up a ticket. 

On the specific refactoring: While conceptionaly ok, it does not work. You never set or configure the resolver and worse, you create the impression it is actually something that a) can be configured and b) a user might even use the new interface as SPI, stuffing in their own resolver. This is quite the opposite about what we want here.

Last but not least, it does not compile:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (java-compile) on project spring-data-neo4j: Compilation failure
[ERROR] /Users/msimons/Projects/spring-projects/spring-data-neo4j/src/main/java/org/springframework/data/neo4j/core/RelationshipHandler.java:[120,50] error: incompatible types: Collection<Object> cannot be converted to List<?>
```

PR #2837, opened 2023-11-28
build: Allow enforcement to reuse a container.

michael-simons:
This orders reusable containers on wish and is meant for scenarios in which a CI does not allow "native" use of reusable test containers but will make sure to clean them up after CI ran. Export a variable `SDN_FORCE_REUSE_OF_CONTAINERS` as `true` and SDN tests won't close the container, regardless whether the underlying Testcontainers supports reuse or not.

### DIFF
--- a/src/test/java/org/springframework/data/neo4j/test/Neo4jExtension.java
+++ b/src/test/java/org/springframework/data/neo4j/test/Neo4jExtension.java
@@ -38,12 +38,9 @@ import org.testcontainers.utility.TestcontainersConfiguration;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


michael-simons:
@markpollack I am not to keen on touching the Jenkinsfile, but if you would please add an environment variable named `SDN_FORCE_REUSE_OF_CONTAINERS` with the value of `true`, things should improve.

mp911de:
From (average):

```
Tests run: 19, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 32.447
```

To (average):

```
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.601 s
```





That is a massive improvement to run the build in 1/5th of the original time. In the unlikely event of nerd sniping, as the driver is netty-base, the driver startup could be optimized by creating an `EventLoopGroup` instance with daemon threads, holding on to it throughout the tests and using `DriverFactory` directly. Just a random thought.

In any case, thanks a lot for the change, that's merged now and brings us down from roughly an hour to 20 minutes.

PR #2838, opened 2023-11-28
build: Keep eventloop around.

michael-simons:
Uses the internal factory method (which actually exists for that purpose, core product uses it too) and keep an event loop group around.


mp911de:
Thank you for your contribution. That's merged and backported now. That took away another 30 seconds from the build time 👍 .

michael-simons:
Neat! Same on local.

PR #2870, opened 2024-03-11
Skip entity detection for converted properties

### LINKED ISSUE #2869: Refine `DefaultNeo4jPersistentProperty.isEntity()` and `getPersistentEntityTypeInformation()` checks
Based on https://github.com/spring-projects/spring-data-commons/issues/3056, type checks to include a property's type as entity require refinement to check against a registered converter since the entity detection mechanism in `BasicPersistentProperty.getPersistentEntityTypeInformation()` no longer uses `isEntity()` due to internal recursive calls.

mp911de:
DefaultNeo4jPersistentProperty.isEntity() and getPersistentEntityTypeInformation() now return that the property does not map to an entity if a converter is registered.

Closes #2869

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-neo4j</artifactId>
-	<version>7.3.0-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


meistermeier:
merged with caddcb636 and a1f456665

PR #2895, opened 2024-04-12
refactor: Move away from deprecated Cypher-DSL constructs.

michael-simons:
All relevant Cypher integrations is available on `Cypher` for a while now.
To be able to remove them in Cypher-DSL 2024.0.0, we must make sure they are not used here anymore, as Cypher-DSL has a couple of integration tests with SDN6.
Also, leading by example is always nice.


meistermeier (approved):
Testing (also with the branch of CypherDSL that includes the removals of the deprecated API) confirms that all works as before ✅ 
A scan for deprecation shows only the known things around `id`/`internalId` that we just cannot fix while still supporting those identity types.

Issue #2944, opened 2024-08-27
RetryExceptionPredicate has the wrong error messages


GregDThomas:
`org.springframework.data.neo4j.core.support.RetryExceptionPredicate` contains two exception messages that are retryable;

```java
private static final Set<String> RETRYABLE_ILLEGAL_STATE_MESSAGES = Set.of(
			"Transaction must be open, but has already been closed.",
			"Session must be open, but has already been closed.");
```

Neither of these messages appear anywhere in the code base. However, very similar messages (without a trailing dot) appear in exceptions raised in `org.springframework.data.neo4j.core.transaction.Neo4jTransactionHolder`. Given these exceptions should be retryable, these exception messages should match.

(Background; we've been wondering why some exceptions weren't being retried in our code base - finally narrowed it down to this!)

GregDThomas:
I'm guessing this could/should be backported to 7.3.x, maybe 7.2.x depending on when the change was made?

michael-simons:
I will take care of this for all supported branches. :)

GregDThomas:
Thank you!

PR #2945, opened 2024-08-27
Fix issue #2944 by making the retryable exception messages consistent

GregDThomas:
- [ ] You have read the [Spring Data Neo4j contribution guidelines](https://github.com/spring-projects/spring-data-neo4j/blob/master/CONTRIBUTING.adoc).
- [ ] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [ ] You submit test cases (unit or integration tests) that back your changes.
- [ ] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).


michael-simons (approved):
You have been faster than me, exactly the way I would have fixed it. Thank you.

PR #2962, opened 2024-10-18
issue/2953

### LINKED ISSUE #2953: Neo4jTemplate.DefaultExecutableQuery.getSingleResult() always executes in non-readonly transaction
When activating `Neo4jTransactionManager.setValidateExistingTransaction(true)` a simple `findById()` will result in an `IllegalStateException`. This is due to `getSingleResult()` always executing in the default transactional-template whereas the outer call `findById()` executes in a readonly-transaction.

How to reproduce:

1. activate transaction validation as described in @michael-simons tips (see below)
2. call `findById()` on a neo4j-template or -repository 

See also
- https://info.michael-simons.eu/2018/09/25/validate-nested-transaction-settings-with-spring-and-spring-boot/

mkoertgen:
fixes #2953

### DIFF
--- a/src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
+++ b/src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
@@ -216,7 +216,7 @@ public final class Neo4jTemplate implements
 		return transactionTemplateReadOnly.execute(tx -> {
 			PreparedQuery<Long> preparedQuery = PreparedQuery.queryFor(Long.class).withCypherQuery(cypherQuery)
 					.withParameters(parameters).build();
-			return toExecutableQuery(preparedQuery).getRequiredSingleResult();
+			return toExecutableQuery(preparedQuery, true).getRequiredSingleResult();
 		});
 	}
 
@@ -230,7 +230,8 @@ public final class Neo4jTemplate implements
 		return transactionTemplateReadOnly
 				.execute(tx -> {
 					Neo4jPersistentEntity<?> entityMetaData = neo4jMappingContext.getRequiredPersistentEntity(domainType);
-					return createExecutableQuery(domainType, resultType, QueryFragmentsAndParameters.forFindAll(entityMetaData))
+					return createExecutableQuery(
+							domainType, resultType, QueryFragmentsAndParameters.forFindAll(entityMetaData), true)
 							.getResults();
 				});
 	}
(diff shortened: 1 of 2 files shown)


meistermeier:
Thanks a lot. I hope to have a look later this week (🤞 )

src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
```diff
@@ -1191,26 +1198,31 @@ public void setTransactionManager(@Nullable PlatformTransactionManager transacti
...
+		return toQuery(preparedQuery, readOnly);
 	}
 
 	@Override
 	public <T> ExecutableQuery<T> toExecutableQuery(PreparedQuery<T> preparedQuery) {
+        return toQuery(preparedQuery, false);
+	}
 
-		return new DefaultExecutableQuery<>(preparedQuery);
+	private <T> ExecutableQuery<T> toQuery(PreparedQuery<T> preparedQuery, boolean readOnly) {
```
meistermeier, line 1224:
```suggestion
	private <T> ExecutableQuery<T> toExecutableQuery(PreparedQuery<T> preparedQuery, boolean readOnly) {
```

minor detail, but I see no reason why this should not have the same name.

src/main/java/org/springframework/data/neo4j/core/Neo4jTemplate.java
```diff
@@ -744,21 +748,24 @@ public void deleteAll(Class<?> domainType) {
 				});
 	}
 
-	private <T> ExecutableQuery<T> createExecutableQuery(Class<T> domainType, Statement statement) {
-		return createExecutableQuery(domainType, null, statement, Collections.emptyMap());
+	private <T> ExecutableQuery<T> createExecutableQuery(Class<T> domainType, Statement statement, boolean readOnly) {
```
meistermeier, line 751: Could we place name it like `readOnlyTransaction` or similar? Of course it makes sense in the current context of the transaction PR, but I know that in a few weeks, I need to think a second more about what this parameter stands for.

meistermeier:
Thanks again for this PR. I think it makes sense in this way, even though I am not completely happy with the syntax of `readOnlyTxManager....(...., readOnly=true)`. It feels somehow like a duplication. On the other hand I see no way how the write-mode transaction manager in the executable query should know that it is about to join an ongoing transaction.

Something that is missing in the PR is the support for the reactive template manager (you're missing all the fun ;) )
Unfortunately there is no validate method on the reactive transaction manager that will help us (you) to verify that this is working, or I cannot find it right now.

mkoertgen:
Hi @meistermeier ,

Will add the reactive part. Thanks for reviewing 👍

mkoertgen:
Hi @meistermeier ,

As of now, [AbstractReactiveTransactionManager](https://github.com/spring-projects/spring-framework/blob/main/spring-tx/src/main/java/org/springframework/transaction/reactive/AbstractReactiveTransactionManager.java) does not support transactional validation as does [AbstractPlatformTransactionManager.setValidateExistingTransaction](https://github.com/spring-projects/spring-framework/blob/e57464dac83ed60e3146b3360bb66dc8d29e5551/spring-tx/src/main/java/org/springframework/transaction/support/AbstractPlatformTransactionManager.java#L242).

So it does not make sense to add it now.

meistermeier:
Sorry to come back to this that late. Had a longer look at this today and I am happy with this change. 
Could you please rebase/squash the commits and add the Sign-Off? (https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring)

mkoertgen:
No problem. Will do 👍

mkoertgen:
There you go @meistermeier.

PR #3009, opened 2025-04-28
Add Vector type support

meistermeier:
Map `org.springframework.data.domain.Vector` to `setNodeVectorProperty` procedure call.

Open documentation tasks

- [x] `double` vector in SDN only
- [x] precision on rest/representation
- [x] only one vector property allowed for an entity definition

Note to myself, there should be a close the issue commit on squash.

### DIFF
--- a/src/main/antora/modules/ROOT/pages/appendix/conversions.adoc
+++ b/src/main/antora/modules/ROOT/pages/appendix/conversions.adoc
@@ -188,6 +188,10 @@ If you require the time zone, use a type that supports it (i.e. `ZoneDateTime`) 
 |Point with CRS 4326 and x/y corresponding to lat/long
 |
 
+|`org.springframework.data.domain.Vector`
+|persisted through `setNodeVectorProperty`
+|
+
 |Instances of `Enum`
 |String (The name value of the enum)
 |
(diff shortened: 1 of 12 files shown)


src/main/antora/modules/ROOT/pages/appendix/conversions.adoc
```diff
@@ -210,6 +214,18 @@ If you require the time zone, use a type that supports it (i.e. `ZoneDateTime`)
 
 |===
 
+[[build-in.conversions.vector]]
+=== Vector type
+Spring Data has its own type for vector representation `org.springframework.data.domain.Vector`.
+While this can be used as a wrapper around a `float` or `double` array, Spring Data Neo4j supports only the `double` variant right now.
+From a user perspective, it is possible to only define the `Vector` interface on the property definition and use either `double` or `float`.
+But on the database-side those values will be persisted as `doubles` for the time being to support manual persisted arrays and
+values persisted through `setNodeVectorProperty`.
```
michael-simons, line 223: Bitte so:

Neo4j will store both `double` and `float` variants as a 64bit Cypher `FLOAT` value, which is consistent with values persisted through Cypher and the dedicated `setNodeVectorProperty` function that Spring Data Neo4j uses to persist the property.

michael-simons (approved):

meistermeier:
Merged via
03c4628dc (8.0.x)
bbf261e75 (main)

PR #3017, opened 2025-06-04
chore: Align formatting with Spring Framework and Spring Boot.

michael-simons:
We use the Spring Formatter already in many Neo4j projects and are quite happy with it.

It feels natural to use it with a Spring project, too. We take the opportunity to apply the formatter to the upcoming 8.0.x release. We think that the now increased difficulty of back porting things to the 7.x lines is outweighed by the better format and the simpler tooling.


meistermeier (approved):
YOLO, right?

PR #3122, opened 2026-09-13
Fix @CompositeProperty converter matching for parameterized property types

### LINKED ISSUE #3035: @CompositeProperty converter rejected when entity field uses parameterized generic type
When an entity field is declared as a parameterized type (e.g. `LocalizedValue<String>`), a matching `Neo4jPersistentPropertyToMapConverter<String, LocalizedValue<String>>` is rejected with:

```java
The property type [...].LocalizedValue<java.lang.String> created by [...].LocalizedValueMapConverter
used on name in [...].OrganizationUnit doesn't match the actual property type
```

If the converter is changed to use the raw type (LocalizedValue), it works.

**Versions**
•	Spring Data Neo4j: 7.5.2
•	Spring Data Commons: 3.5.2
•	Spring Data REST: 4.5.2
•	Spring Boot: 3.5.4
•	Neo4j Java Driver: 5.28.9
•	Java: 21

### Minimal example

**Domain value type**

package com.example.localized;

```java
public class LocalizedValue<T> {
  // internally holds Map<Locale,T>; getters omitted for brevity
}
```

**Converter (fails)**

```java
public class LocalizedValueMapConverter
    implements Neo4jPersistentPropertyToMapConverter<String, LocalizedValue<String>> {

  @Override public Map<String, Value> decompose(LocalizedValue<String> lv, Neo4jConversionService cs) { ... }
  @Override public LocalizedValue<String> compose(Map<String, Value> src, Neo4jConversionService cs) { ... }
}
```

**Entity**

```java
@Node
public class OrganizationUnit {
  @Id @GeneratedValue Long id;

  @CompositeProperty(converter = LocalizedValueMapConverter.class, prefix = "name")
  private LocalizedValue<String> name;

  // getters/setters use the same parameterized type
}
```

### Observed

- Throws `IllegalArgumentException` above.
- Neo4jMappingContext appears to resolve the property type as raw `LocalizedValue`, so the parameterized converter is considered a mismatch.
- I tried a Converter with `Neo4jPersistentPropertyToMapConverter<String, LocalizedValue<T>>` aswell, same issue.

### Workarounds
•	Make the converter raw:

```java
class LocalizedValueMapConverter
    implements Neo4jPersistentPropertyToMapConverter<String, LocalizedValue> { ... }
```

(Works, but loses compile-time type safety.)

### Expected

- A converter targeting `LocalizedValue<String>` should match a field declared as `LocalizedValue<String>`.

Happy to provide a tiny reproducer repo if helpful.

arimu1:
## Summary

- Fix `CompositePropertyConverterFactory` so `@CompositeProperty` converters whose `Neo4jPersistentPropertyToMapConverter` type parameter is a parameterized generic (e.g. `LocalizedValue<String>`) match entity fields declared with the same type.
- Previously validation compared the field’s raw erasure (`getActualType()`) to the converter’s resolved generic type, causing `IllegalArgumentException` for valid configurations (gh-3035).
- Regression coverage is in `Neo4jMappingContextTests#shouldAcceptCompositeConverterWithParameterizedPropertyType` (inner `WithParameterizedCompositeUsage` / `ParameterizedValue` fixtures). Unused gh-3035 integration package removed — nothing referenced those types.

## Test plan

- [x] `./mvnw -Dtest=Neo4jMappingContextTests#shouldAcceptCompositeConverterWithParameterizedPropertyType test`
- [x] `./mvnw '-Dtest=Neo4jMappingContextTests#should*Composite*' test` (6 tests)

Closes #3035


meistermeier:
Thanks for your contribution

